PR #1, opened 2020-11-13
Offer youtube-dlc as an alternative to youtube-dl

peppizza:
https://github.com/serenity-rs/serenity/pull/1076

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,7 +107,6 @@ default = [
     "serenity-rustls",
     "driver",
     "gateway",
-    "youtube-dl"
 ]
 gateway = [
     "flume",
@@ -136,7 +135,6 @@ driver = [
     "url",
     "xsalsa20poly1305",
 ]
-youtube-dl = []
 youtube-dlc = []
 rustls = ["async-tungstenite/tokio-rustls"]
 native = ["async-tungstenite/tokio-native-tls"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,6 +107,7 @@ default = [
     "serenity-rustls",
     "driver",
     "gateway",
+    "youtube-dl"
 ]
 gateway = [
     "flume",
@@ -135,6 +136,8 @@ driver = [
     "url",
     "xsalsa20poly1305",
 ]
+youtube-dl = []
+youtube-dlc = []
 rustls = ["async-tungstenite/tokio-rustls"]
 native = ["async-tungstenite/tokio-native-tls"]
 serenity-rustls = ["serenity/rustls_backend", "rustls", "gateway", "serenity-deps"]
--- a/build.rs
+++ b/build.rs
@@ -20,4 +20,13 @@ compile_error!(
     If you are unsure, go with `stock-zlib`."
 );
 
+#[cfg(all(feature = "youtube-dl", feature = "youtube-dlc"))]
+compile_error!(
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -135,6 +136,8 @@ driver = [
     "url",
     "xsalsa20poly1305",
 ]
+youtube-dl = []
+youtube-dlc = []
```
FelixMcFelix, line 138: I think just `youtube-dlc` will suffice: see below.

build.rs
```diff
@@ -20,4 +20,13 @@ compile_error!(
...
 
+#[cfg(all(feature = "youtube-dl", feature = "youtube-dlc"))]
+compile_error!(
+    "Only youtube-dl or youtube-dlc can be selected \
+    - `youtube-dl` is the standard downloading tool of youtube videos \
+    - `youtube-dlc` is a fork of youtube-dl that aims to solve bugs with youtube-dl \
+    If you are unsure, go with `youtube-dl`, however, if you encounter any errors with `youtube-dl`,
+    such as tracks ending immedietly when they are queued/played, try 'youtube-dlc'."
+);
+
```
FelixMcFelix: If you add only one new feature (`youtube-dl`), this check can be avoided.

In general, it's better to avoid adding in feature traps/incompatibilities where you can.

src/input/ytdl_src.rs
```diff
@@ -14,7 +14,13 @@ use std::{
 use tokio::task;
 use tracing::trace;
 
+#[cfg(feature = "youtube-dl")]
+const YOUTUBE_DL_COMMAND: &str = "youtube-dl";
+#[cfg(feature = "youtube-dlc")]
+const YOUTUBE_DL_COMMAND: &str = "youtube-dlc";
```
FelixMcFelix: Can you make this so that the default is `"youtube-dl"`, and the `youtube-dlc` feature changes this?

E.g., constant computation allows constructs such as:
```rust
const TEST: &str = if cfg!(feature = "") {
    "A"
} else {
    "B"
};
```

src/input/ytdl_src.rs
```diff
@@ -14,7 +14,13 @@ use std::{
...
+
 /// Creates a streamed audio source with `youtube-dl` and `ffmpeg`.
+#[cfg(any(feature = "youtube-dl", feature = "youtube-dlc"))]
```
FelixMcFelix: With one override feature, this check is unnecessary. This also stops `Restartable` sources from failing to compile.

src/input/ytdl_src.rs
```diff
@@ -102,6 +108,7 @@ pub(crate) async fn _ytdl(uri: &str, pre_args: &[&str]) -> Result<Input> {
 
 /// Creates a streamed audio source from YouTube search results with `youtube-dl`,`ffmpeg`, and `ytsearch`.
 /// Takes the first video listed from the YouTube search.
+#[cfg(any(feature = "youtube-dl", feature = "youtube-dlc"))]
```
FelixMcFelix: See the previous instance,

FelixMcFelix (changes requested):
Thanks for the PR, it'll be useful to support this as a drop in.

I've made some comments around what feature flags to use, and how -- mainly to minimise friction.

FelixMcFelix (approved):

FelixMcFelix:
Lgtm. Thanks!

PR #16, opened 2020-11-18
Add function to return a vector of all the handles in a queue

peppizza:
Basically, an extension to current() but returns a vector of all the handles in a queue

### DIFF
--- a/src/tracks/queue.rs
+++ b/src/tracks/queue.rs
@@ -114,7 +114,7 @@ impl EventHandler for QueueHandler {
                 let ended_uuid = ts.first().map(|handle| handle.1.uuid());
 
                 queue_uuid.is_some() && queue_uuid == ended_uuid
-            }
+            },
             _ => false,
         };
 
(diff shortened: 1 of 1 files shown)


src/tracks/queue.rs
```diff
@@ -278,6 +278,13 @@ impl TrackQueue {
 
         inner.stop_current()
     }
+
+    /// Returns Vec of all handles in queue
```
FelixMcFelix: Main points to mention in docs after the headline:
* Reordering the queue can't be done here, and has to go through `modify_queue`. The returned Vec is a snapshot of the queue at the time of call, and might change afterwards.
* Calling `stop` on a handle will effectively skip that track.

Minor nits:
* can you please end doc comments with a full stop?
* Maybe something like "Returns a list of currently queued tracks." for the headline?

src/tracks/queue.rs
```diff
@@ -278,6 +278,13 @@ impl TrackQueue {
...
+    pub fn to_vec(&self) -> Vec<TrackHandle> {
```
FelixMcFelix:
```suggestion
    pub fn current_queue(&self) -> Vec<TrackHandle> {
```

This is a little more descriptive of the function's purpose here.

FelixMcFelix (changes requested):
Thanks for the pull request. This should be a good convenience method.

A few nits, mainly in naming and pointing out some gotchas in documentation for future users.

FelixMcFelix (approved):

PR #18, opened 2020-11-23
Fixed serenity dependency default features

Maspenguin:
This change is needed to fix this issue in serenity when using songbird.
```
thread 'main' panicked at 'The `framework`-feature is enabled (it's on by default), but no framework was provided.
If you don't want to use the command framework, disable default features and specify all features you want to use.'
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,7 @@ version = "0.7"
 
 [dependencies.serenity]
 optional = true
+default-features = false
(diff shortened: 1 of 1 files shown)


FelixMcFelix:
Lgtm, I hadn't tested frameworkless serenity, so this is a good catch. Thanks!

arqunis (approved):

PR #25, opened 2020-12-03
add metadata field to TrackHandle

### LINKED ISSUE #24: Track metadata in TrackHandle
Being able to access track metadata from TrackHandle struct would be very useful, like in use cases such as listing queue or transitioning from one song to another.

peppizza:
Closes #24.  Untested but should work theoretically.  Does not solve the issue of storing extra data into the handle

### DIFF
--- a/src/tracks/handle.rs
+++ b/src/tracks/handle.rs
@@ -167,7 +167,13 @@ impl TrackHandle {
         self.uuid
     }
 
-    /// Returns the metadata stored in the handle
+    /// Returns the metadata stored in the handle.
+    ///
+    /// Metadata is cloned from the inner [`Input`] at
+    /// the time a track/handle is created, and is effectively
+    /// read-only from then on.
+    ///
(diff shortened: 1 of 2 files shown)


src/tracks/handle.rs
```diff
@@ -157,6 +167,11 @@ impl TrackHandle {
         self.uuid
     }
 
+    /// Returns the metadata stored in the handle
```
FelixMcFelix:
```suggestion
    /// Returns the metadata stored in the handle.
    ///
    /// Metadata is cloned from the inner [`Input`] at
    /// the time a track/handle is created, and is effectively
    /// read-only from then on.
    ///
    /// [`Input`]: crate::input::Input
```

FelixMcFelix (changes requested):
Thanks for the useful change, just an addendum to docs and then it should be all good to go!

FelixMcFelix (approved):

FelixMcFelix:
Thanks!

Issue #17, opened 2020-11-19
Add Typemap to TrackState


peppizza:
I think it would be useful to allow users of the library to store data alongside the handle in a TrackQueue, such as storing the metadata of the song, to allow for easier retrieval, instead of having to manage 2 queues at once.

peppizza:
It might actually be better to store a data field in TrackState, as I assume thats what it was designed to do

FelixMcFelix:
I had considered placing a TypeMap onto each Driver (and the track queue into that) as a more general way of offering extra per-driver state, but I think this needs some development to be feasible (i.e., some way to specify a closure to add default entries and so on).

I wouldn't say that TrackState is the right place for this. That is intended solely for state which the *driver* will regularly change. I think it'd be reasonable to use the new uuids to store and access metadata via a hashmap. This is a little heavyweight right now. given you'll need to use track end events to clear it out, but I think it might be useful to store `Arc<Metadata>` in Inputs and spawned handles down the line to make this easier.

Similarly, TrackQueue is meant to be pretty simple; more as a proof-of-concept to show how users can build their own complexity on the event system. I'm wary of dumping too much resource into it for now.

peppizza:
I would be ok with having a store for the metadata instead of a full blown typemap as I think storing metadata along with tracks will be a very common use case

peppizza:
After thinking about it, instead of using a TypeMap, it might be better to only allow inserting of one user-defined struct into the trackhandle

FelixMcFelix:
Closed by d42e09f.

PR #28, opened 2020-12-06
Added source_url and thumbnail fields to Metadata

saanuregh:
Some useful changes to Metadata.
- Metadata: Added `source_url` and `thumbnail` fields and additional changes to `from_ytdl_output` fn.

### DIFF
--- a/examples/serenity/voice/src/main.rs
+++ b/examples/serenity/voice/src/main.rs
@@ -227,7 +227,7 @@ async fn play(ctx: &Context, msg: &Message, mut args: Args) -> CommandResult {
             },
         };
 
-        handler.play_source(source, Some(msg.author.id));
+        handler.play_source(source);
 
         check_msg(msg.channel_id.say(&ctx.http, "Playing song").await);
     } else {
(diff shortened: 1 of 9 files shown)


FelixMcFelix:
Hi Saanu, thanks for the pull request.

The metadata changes are fine, sensible, and useful. Those look good to me.

I disagree with the requester IDs change, though. This seems to be very queue-specific functionality, and has a large API surface impact (i.e., users who don't care about queues now have an extra field added to all their method calls). I think that it's reasonable to expect that devs keep track of this themselves until something like #17 is cleanly implemented/solved. The reason why it should wait for something like this is that the requester ID *is* metadata, in a sense -- but it shouldn't affect anyone who isn't using the demo queue. The library is more general-purpose than that.

In future, please try to keep separate feature additions as separate PRs where possible! :)

saanuregh:
That makes sense. I have reverted Requester ID changes in Tracks.
Until #17 this can be implemented what if the requester field is added to metadata, which can be easily mutated if wanted?

FelixMcFelix:
I think the requester field steps over the line from "intrinsic metadata" (i.e., data about the input itself) to something more use-specific, so I would say not for now.

FelixMcFelix (approved):

PR #33, opened 2020-12-28
Add set_uuid method to track and trackhandle

peppizza:
Adds methods to update the tracks and handles uuid if the user needs it tied to something else before the tracks creation

### DIFF
--- a/src/driver/connection/error.rs
+++ b/src/driver/connection/error.rs
@@ -93,9 +93,8 @@ impl fmt::Display for Error {
             CryptoModeUnavailable => write!(f, "server did not offer chosen encryption mode."),
             EndpointUrl => write!(f, "endpoint URL received from gateway was invalid."),
             ExpectedHandshake => write!(f, "voice initialisation protocol was violated."),
-            IllegalDiscoveryResponse => {
-                write!(f, "IP discovery/NAT punching response was invalid.")
-            },
+            IllegalDiscoveryResponse =>
+                write!(f, "IP discovery/NAT punching response was invalid."),
             IllegalIp => write!(f, "IP discovery/NAT punching response had bad IP value."),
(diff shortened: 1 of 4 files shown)


FelixMcFelix:
Hi, thanks for the PR!

I think the idea of what you want to make doable is good (have `Track`s with custom UUIDs). However, what I'm trying to go for right now is to have all of the data in a handle placed *behind* the `Arc<...>`, because this is going to greatly simplify sticking a `TypeMap` in there and it ensures that cloning handles is very cheap.

The problem is that keeping the UUID behind the Arc makes this quite tricky. To get around this, I can suggest a few ways:

1. Break [`create_player`](https://serenity-rs.github.io/songbird/current/songbird/tracks/fn.create_player.html) out so that you can `impl From<Input> for Track` -- you might also want to call this `Track::new(input: Input) -> Self`. You need this to only have one strong reference to the `InnerHandle`, which lets you [modify its contents](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut). What you can do then is change `set_uuid` to return a `bool` or `Option` according to whether the UUID *could* be changed.
2. Add a `Track::new_with_uuid(input, uuid)` which constructs a handle with a non-random UUID.

These should allow custom UUIDs without letting a user attempt to do something invalid (like change the UUID of a handle after it's been passed to the driver). It's useful to have constant state remain consistent and cheap to access/move around.

As an aside, make sure to format with `cargo +nightly fmt`, as some of the configurations we're using are unstable (such as "match_arm_blocks").

peppizza:
I created a new method create_player_with_uuid to construct the track and handle with a custom uuid

FelixMcFelix (approved):
Looks good, thanks very much!

PR #40, opened 2021-02-02
Library: Add compatibility for legacy Tokio 0.2

FelixMcFelix:
Adds support to the library for tokio 0.2 backward-compatibility. This should hopefully benefit users blocked on other libraries, and prevent lavalink-rs from being blocked on this feature.

These can be reached using, e.g., `gateway-tokio-02`, `driver-tokio-02`, `serenity-rustls-tokio-02`, and `serenity-native-tokio-02` features.

Naturally, this requires some jiggering about with features and the underlying CI, which has been taken care of. Twilight can't be handled in this way, as their last tokio 0.2 version uses the deprecated Discord Gateway v6.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       uses: actions-rs/clippy-check@v1
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
-        args: --all-features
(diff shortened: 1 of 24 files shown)


FelixMcFelix:
Double checked that this was working with serenity, seems good to go.

PR #41, opened 2021-02-07
removing unnecessary methods etc

ghost:
and using an error handling similar to the basic ping bot example

### DIFF
--- a/examples/serenity/voice/src/main.rs
+++ b/examples/serenity/voice/src/main.rs
@@ -53,7 +53,7 @@ async fn main() {
 
     let framework = StandardFramework::new()
         .configure(|c| c
-                   .prefix("~"))
+                   .prefix("!"))
         .group(&GENERAL_GROUP);
 
     let mut client = Client::builder(&token)
@@ -63,14 +63,15 @@ async fn main() {
         .await
         .expect("Err creating client");
 
-    let _ = client.start().await.map_err(|why| println!("Client ended: {:?}", why));
+    if let Err(why) = client.start().await {
+        println!("Err starting client: {:?}", why))
+    };
 }
 
 #[command]
 #[only_in(guilds)]
 async fn deafen(ctx: &Context, msg: &Message) -> CommandResult {
-    let guild = msg.guild(&ctx.cache).await.unwrap();
-    let guild_id = guild.id;
+    let guild_id = msg.guild_id;
 
     let manager = songbird::get(ctx).await
         .expect("Songbird Voice client placed in at initialisation.").clone();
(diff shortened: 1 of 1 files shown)


examples/serenity/voice/src/main.rs
```diff
@@ -63,14 +63,15 @@ async fn main() {
...
+        println!("Err starting client: {:?}", why))
+    };
 }
 
 #[command]
 #[only_in(guilds)]
 async fn deafen(ctx: &Context, msg: &Message) -> CommandResult {
-    let guild = msg.guild(&ctx.cache).await.unwrap();
-    let guild_id = guild.id;
+    let guild_id = msg.guild_id;
```
arqunis, line 74: `Message::guild_id` can be `None` if the message came from a DM, but in this instance it is guarded against with the `#[only_in(guilds)]` precondition check, so this should be `unwrap`ed:
```suggestion
    let guild_id = msg.guild_id.unwrap();
```

examples/serenity/voice/src/main.rs
```diff
@@ -129,8 +130,7 @@ async fn join(ctx: &Context, msg: &Message) -> CommandResult {
 #[command]
 #[only_in(guilds)]
 async fn leave(ctx: &Context, msg: &Message) -> CommandResult {
-    let guild = msg.guild(&ctx.cache).await.unwrap();
-    let guild_id = guild.id;
+    let guild_id = msg.guild_id;
```
arqunis, line 133: Ditto.

examples/serenity/voice/src/main.rs
```diff
@@ -53,7 +53,7 @@ async fn main() {
 
     let framework = StandardFramework::new()
         .configure(|c| c
-                   .prefix("~"))
+                   .prefix("!"))
```
arqunis, line 56: Why the change in prefix? All of Serenity's examples use the tilde (`~`) as a prefix, and `!` is too commonly used. This is going to cause conflicts with other bots if a user were to run this example mostly unchanged.

arqunis (changes requested):

examples/serenity/voice/src/main.rs
```diff
@@ -63,14 +63,15 @@ async fn main() {
         .await
         .expect("Err creating client");
 
-    let _ = client.start().await.map_err(|why| println!("Client ended: {:?}", why));
+    if let Err(why) = client.start().await {
+        println!("Err starting client: {:?}", why))
```
FelixMcFelix, line 67: Aside from Acdenis's comments, you've put in an extra bracket here. The example will not compile as-is.

examples/serenity/voice/src/main.rs
ghost: isnt there any other way guild_id is none? maybe some bug i dont think it requires panicking so i'd just return if its None

examples/serenity/voice/src/main.rs
ghost: it doesnt have to be ! it just seemed weird since ~ is already the default

examples/serenity/voice/src/main.rs
ghost: yes i should learn to use vscode for prs sorry

FelixMcFelix:
I'm closing this for now as there's been no activity in over a month. Feel free to reopen if you return to this!

PR #45, opened 2021-03-13
Fix input source timestamp pre-input argument decimal formatting

DasEtwas:
The previous code would have formatted timestamps such as `3.009` seconds as `3.9`.


FelixMcFelix (approved):
Thanks for the pull request; good catch! Looks good to me.

Issue #56, opened 2021-04-09
Calling `.leave()` during audio playing blocks audio playing on reconnection


JellyWX:
When calling `.leave()` on a Call, if that call is already playing audio and the `.leave()` interrupts it, then upon rejoining the call, the bot will not play audio.

Calling `.remove()` instead to drop the call is fine

FelixMcFelix:
I updated my main bot to 0.1.5 (since I was awaiting a serenity version update) and have quickly repro'd and confirmed. I have a feeling this was introduced by #46.

DoumanAsh:
I encountered similar situation with `remove`
If I call `Call::stop` and then `Songbird::remove`  ~and then attempt to reconnect using `Songbird::join`~
```
thread '<unnamed>' panicked at 'Shouldn't be mixing packets without access to a cipher + UDP dest.', C:\Users\Douman\data\Soft\Dev\rustup\cargo_home\registry\src\github.com-1ecc6299db9ec823\songbird-0.1.5\src\driver\tasks\mixer.rs:492:14
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0\/library\std\src\panicking.rs:493
   1: core::panicking::panic_fmt
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0\/library\core\src\panicking.rs:92
   2: core::option::expect_failed
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0\/library\core\src\option.rs:1292
   3: core::option::Option<mut songbird::driver::tasks::message::mixer::MixerConnection*>::expect<mut songbird::driver::tasks::message::mixer::MixerConnection*>
             at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0\library\core\src\option.rs:349
   4: songbird::driver::tasks::mixer::Mixer::prep_and_send_packet
             at C:\Users\Douman\data\Soft\Dev\rustup\cargo_home\registry\src\github.com-1ecc6299db9ec823\songbird-0.1.5\src\driver\tasks\mixer.rs:489
   5: songbird::driver::tasks::mixer::Mixer::cycle
             at C:\Users\Douman\data\Soft\Dev\rustup\cargo_home\registry\src\github.com-1ecc6299db9ec823\songbird-0.1.5\src\driver\tasks\mixer.rs:478
[...]

PR #60, opened 2021-04-10
Introduce Call::current_channel

### LINKED ISSUE #58: Call::connection should containt channel id to determine if you're already in channel
Subj.
It would be useful to know if you're already in channel or need to join new one

DoumanAsh:
I'm not sure how you want to organize connection info and whether channel id should be separate from it.

For now I left it separated, just getting API to access it for user so that I could know whether I need to join new one or not

Fixes #58

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -241,6 +241,10 @@ impl Call {
     }
 
     /// Returns `id` of the channel, if connected to any.
+    ///
+    /// **Note:**: Returned `id` is of the channel, to which bot performed connection.
+    /// It is possible that it is different from actual channel due to ability of server's admin to
+    /// move bot from channel to channel. This is to be fixed with next breaking change release.
     #[instrument(skip(self))]
     pub fn current_channel(&self) -> Option<ChannelId> {
         match &self.connection {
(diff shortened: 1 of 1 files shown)


src/handler.rs
```diff
@@ -240,6 +240,15 @@ impl Call {
         }
     }
 
+    /// Returns `id` of the channel, if connected to any.
```
FelixMcFelix, line 243:
```suggestion
    /// Returns `id` of the channel, if connected to any.
    ///
    /// **Note**: The returned `id` will not correctly update if a connected bot
    /// is moved between channels manually. This will be fixed in the next
    // breaking release.
```

FelixMcFelix (changes requested):
Thanks for the PR! This is useful, and can be modified easily to account for #47's `ConnectionInfo` changes when it's time to release that.

Just need an additional comment on a known bug until next breaking due to #47 (since this PR will go into a patch release after #56 is fixed). Apart from that, lgtm!

src/handler.rs
DoumanAsh: I expanded a bit your suggestion. Check it out

FelixMcFelix (approved):
I think that expands the explanation nicely, thanks!

PR #73, opened 2021-05-18
Impl StdError

### LINKED ISSUE #72: Impl std::error::Error for songbird error types
This requires that every error impl Display.

vilgotf:
Closes #72

Most errors seem to already impl std::error::Error so this was a smaller change than expected.

### DIFF
--- a/src/input/error.rs
+++ b/src/input/error.rs
@@ -69,6 +69,48 @@ impl From<OpusError> for Error {
     }
 }
 
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Error::Dca(e) => write!(f, "{}", e),
+            Error::Io(e) => write!(f, "{}", e),
+            Error::Json {
+                error,
+                parsed_text: _,
+            } => write!(f, "{}", error),
+            Error::Opus(e) => write!(f, "{}", e),
+            Error::Metadata => write!(f, "Failed to extract metadata"),
+            Error::Stdout => write!(f, "Failed to create stdout"),
+            Error::Streams => write!(f, "Error while checking if path is stereo"),
+            Error::Streamcatcher(e) => write!(f, "{}", e),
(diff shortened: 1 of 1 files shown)


src/input/error.rs
```diff
@@ -68,6 +69,48 @@ impl From<OpusError> for Error {
...
+            Error::Streamcatcher(e) => write!(f, "{}", e),
+            Error::YouTubeDlProcessing(_) => write!(f, "Processing JSON from youtube-dl failed"),
+            Error::YouTubeDlRun(_) => write!(f, "youtube-dl encountered an error"),
+            Error::YouTubeDlUrl(_) => write!(f, "Missing url field in JSON"),
+        }
+    }
+}
+
+impl StdError for Error {
+    fn cause(&self) -> Option<&dyn StdError> {
```
FelixMcFelix: It looks like `cause` is deprecated: could you please change this to `fn source`?

src/input/error.rs
```diff
@@ -86,6 +129,30 @@ pub enum DcaError {
...
+            DcaError::InvalidHeader => write!(f, "Invalid DCA JSON header"),
+            DcaError::InvalidMetadata(e) => write!(f, "{}", e),
+            DcaError::InvalidSize(e) => write!(f, "Invalid metadata block size: {}", e),
+            DcaError::Opus(e) => write!(f, "{}", e),
+        }
+    }
+}
+
+impl StdError for DcaError {
+    fn cause(&self) -> Option<&dyn StdError> {
```
FelixMcFelix: Same as above.

FelixMcFelix (changes requested):
Thanks very much; I tried to get in a baseline `impl Error` for most structs, but it's clear that some slipped through the gaps.

Just needs some function prototypes changed, as `Error::cause` is now deprecated.

src/input/error.rs
vilgotf: Ah how awkward, I confused the two 😅

vilgotf:
Should I squash or will you?

FelixMcFelix (approved):
LGTM; thanks! I'll squash it here.

PR #74, opened 2021-05-20
Various: Make error messages more idiomatic

vilgotf:
I recently read the `std::error::Error` [trait documentation](https://doc.rust-lang.org/nightly/std/error/trait.Error.html) and found that messages should (in general) be lowercase. After more searching I also found [this](https://users.rust-lang.org/t/do-people-not-care-about-printable-error-chains-a-k-a-how-to-nicely-implement-display-for-an-error/35362) which talks about how `std::error::Error::source` should behave (to produce nice error chains).

This pull requests implements these lessons for *most* error types (I didn't want to touch some of `songbird::error::ConnectionError`'s errors (like `songbird::ws::Error`) due to their complexity).

Tested with `cargo make ready`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -129,6 +129,7 @@ features = ["v4"]
 [dependencies.xsalsa20poly1305]
 optional = true
 version = "0.7"
+features = ["std"]
 
 [dev-dependencies]
 criterion = "0.3"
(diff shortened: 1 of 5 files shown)


vilgotf:
tracing_subscriber dosn't support multiple `Error::source()` OOTB https://github.com/tokio-rs/tracing/issues/1347

FelixMcFelix (approved):
Thanks for the PR, these definitely agree with style/usability guidelines more than before.

PR #75, opened 2021-05-26
Input: Add separate YouTube title and channel to Metadata

vilgotf:
Adds YouTube channel since it's often provided when talking about YouTube content (Rythm includes it when enqueueing a track ([picture](https://user-images.githubusercontent.com/26655508/119718617-113cb700-be68-11eb-85b8-69069580ee37.png))).

Tested with `cargo make ready` and in my own bot

In the future I'd like for there to also be support for who enqueued a track, but I think this info should maybe not be in metadata (since it should probably only contain information about the track).


FelixMcFelix (approved):
Thanks! Looks good!

PR #77, opened 2021-06-13
Tracks: Simplify track end event handler

vilgotf:
Using `?` and `while let Some()` I was able to remove the booleans `front_ended` & `keep_looking`


FelixMcFelix (approved):
Nice logical cleanup/refactor, thanks!

PR #79, opened 2021-06-14
Deps: Bump twilight versions -> 0.5

vilgotf:
Some ugly changes in the example, I'll send in another pull request that will significantly improving it (or should it be included here too?)


FelixMcFelix (approved):
Thanks. I think this looks fine to me, after double checking with twilight's upgrade paths. If you have any other changes in future wrt. readability then those can be added later.

PR #83, opened 2021-06-24
Fix for underflow by clamping

JellyWX:
Clamp the start time to 0 as a potential fix for underflow when playing sounds


FelixMcFelix (approved):
Simple enough fix; thanks very much!

PR #84, opened 2021-06-28
Tracks: Remove box around TrackState

vilgotf:
Boxing is unnecessary since `TrackState` implements copy.

Tested using `cargo make ready`


FelixMcFelix (approved):
Thanks, lgtm. I think there was an earlier version of `TrackState` which had at least one non-`Copy` member, glad this is improved now. :)

PR #87, opened 2021-08-05
Deps: Bump twilight versions -> 0.6

vilgotf:
Includes two more small changes too small to warrant PRs.
1. Removes the `shard_count` parameter from `Songbird::twilight` & `Songbird::twilight_from_config` since the cluster contains it.
2. Drops the `Arc` wrapper around `Songbird` to match against an upcoming twilight 0.7 change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -105,12 +105,12 @@ default-features = false
 
 [dependencies.twilight-gateway]
 optional = true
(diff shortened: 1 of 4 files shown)


FelixMcFelix (approved):
Thanks very much, I appreciate that this also cleans up the Twilight example a little bit!

Issue #67, opened 2021-04-27
Migrate mixing/decoding toolchain to Symphonia


FelixMcFelix:
This issue proposes that we finally move away from (purely) IPC-based decoding to the codec and framing stack in development in [Symphonia](https://github.com/pdeljanov/Symphonia).

This would be an involved piece of work, involving (at least):

* Write an audiopus-based Symphonia codec, and port the DCA framing mechanism. The audiopus-based codec cannot be upstreamed, as Symphonia requires pure-Rust implementations. Similarly, this is not on their current roadmap.
* Implement DASH handling for youtube and other multimedia handlers extracted from youtube-dl.
* Investigate upstreaming MKV support to Symphonia. The MKV muxer underlies formats such as WebM (used primarily by youtube et al.).
* Ensure that Opus packet passthrough can be performed through Symphonia.
* Port our stream-cacheing wrappers to this ecosystem.

There are several gaps here, but it's also an exciting project that should help us leverage their existing (and forthcoming) codec implementations.

FelixMcFelix:
Closed by #89.

PR #89, opened 2021-08-23
Driver/Input: Migrate audio backend to Symphonia

FelixMcFelix:
This begins work on #67, which will be a huge step towards purely in-process audio processing (i.e., one day eliminating ffmpeg).

- [x] Audiopus-based Symphonia codec and DCA1 framing.
- [x] Ensure that Opus packet passthrough can be performed through Symphonia.
- [x] Port mixing infrastructure (and packet passthrough).
- [x] Rework `input` interface.
- [x] Implement basic DASH handling for youtube and other multimedia handlers extracted from youtube-dl.
- [ ] Investigate upstreaming MKV support to Symphonia.
- [x] Port our stream-cacheing wrappers to this ecosystem.

There will be some slight issues stemming from the fact that we will now be parsing metadata from bytestreams as they arrive:
* How will metadata be accessible to users before it enters the mixer? Is this just a `spawn_blocking` on the initial `Format` parse phase? Pass messages out from the main mixer context?
* Users will need to be able to both pass in raw `MediaSourceStream`s, `Read`s, *and* fully processed `(Format, Metadata, ...)` sets -- due to the above, and as some extra metadata will *only* be exposed by youtube-dl, for instance.
* Until MKV support is impl'd, we'll probably need to peek at bytes internally and pass over to ffmpeg on a decode failure. Having such a fallback should probably be toggleable.
* We'll probably need to offer one or two wrappers to hand over `AsyncRead`s into the mixing system (as well as automating extension/MIME type extraction from `File`s or HTTP responses).


FelixMcFelix:
Rough status update: we now have a decent mixing pipeline for audio files of different formats, samples rates, packet lengths etc. into a single buffer. This is horrifically messy. Opus packet passthrough isn't there yet, but should be possible with some care around codec resets. There are some issues around track data cleanup right now.

We also have a wrapper for running seeks/reads over async audio sources (i.e., reqwest calls) and passing those bytes in via ring buffer, which seem to be working great for exposing some amount of flexibility around input sources. In tandem, [mkv support in symphonia seems to be coming along well](https://github.com/pdeljanov/Symphonia/pull/66) -- hopefully with that youtube-dl should integrate nicely to allow e.g. WebMs. I hope that will come down to simple parsing, with occasional renegotiation for a new DASH source if we don't want the stream to cut out halfway through.

FelixMcFelix:
There are a couple things I want to take care of before this is ready:
- [x] A WAV adapter for raw interleaved audio streams. This is probably necessary to support voice bridge use cases since symphonia needs input data to use an actual framing mechanism, e.g., [TS3<->Discord](https://github.com/0xpr03/voice-bridge) and Discord<->Discord.
- [x] Removing the `Option<EventStore>` and `Option<Input>` from the inner `Track` representation, since these are now just wasting space. This might be a good opportunity to also look into a struct-of-arrays layout.
- [x] Some way of inspecting metadata *from the parsed files* by external code. Currently, these can't be moved out of the driver at all, which is awful (but sensible, as a `Format` may append further metadata at any point i the stream in theory).
- [x] Track error events, particularly since seek results are a little trickier to reason about.
- [x] Fix docs and ARCHITECTURE.md to be corrected now that we no longer provide our own input system.

Hopefully afterwards we'll be in a good place for more widespread testing.

FelixMcFelix:
Rearranging `Track`s internally to have a struct-of-arrays layout made no statistical performance difference, aside from making Opus passthrough some 2% worse. For all the effort involved (and the extra mental overhead in making it borrow-checker friendly), this at least justifies the simpler AoS format from a development point of view.

FelixMcFelix:
I think this is now in a very good place in terms of code organisation and API documentation. There are probably some rough edges with the API, but I think these are only going to be sanded down in a pre-release version or similar once people start to actually use it.

One thing I want to do before unmarking this as a draft is to add some `#[cfg(test)]` modifications to the `Driver` to allow it to work headlessly so that we can enforce better integration testing of events, queue systems, and input behaviour. This is really needed, since there's been a lot of internal and external churn.

FelixMcFelix:
I've put a fairly in-depth review and commenting round into the core driver changes, and I've had a more cursory look over the rest of the changes (inputs, adapters, tracks). I'll be merging shortly -- between this and the adventurous use in prod from a handful of users I think this is really quite stable now. Most importantly, it's fast and it's lightweight as a pure Rust implementation should be.

One caveat is that we're waiting on one symphonia bug fix being merged and released before this can go to crates.io, at some point this might need to be pushed out with a disclaimer that the current symphonia fork should be Cargo patched in instead. Hopefully this won't be the case!

PR #90, opened 2021-08-27
Allows yt-dlp usage as another youtube-dl fork

Miezhiko:
Allows yt-dlp usage
https://github.com/yt-dlp/yt-dlp

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -221,6 +221,7 @@ tokio-02-marker = []
 
 # Behaviour altering features.
 youtube-dlc = []
+yt-dlp = []
 builtin-queue = []
 
 # Used for docgen/testing/benchmarking.
--- a/src/input/ytdl_src.rs
+++ b/src/input/ytdl_src.rs
@@ -19,6 +19,8 @@ use tracing::trace;
 
 const YOUTUBE_DL_COMMAND: &str = if cfg!(feature = "youtube-dlc") {
     "youtube-dlc"
+} else if cfg!(feature = "yt-dlp") {
+    "yt-dlp"
 } else {
     "youtube-dl"
 };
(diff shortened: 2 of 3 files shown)


vilgotf:
How about instead of having one feature / yt-dl backend we could expose a compile-time env variable (using `std::option_env!`, default to `youtube-dl` if unset)?

Miezhiko:
for me personally it's more handy to depend with feature, easier than making build script involving env variable for that
```toml
features = [ "yt-dlp"
           , "serenity-rustls"
           , "driver"
           , "gateway"
           ]
```

another possibility is to allow override/provide path to `youtube-dl` app in runtime instead of compile time but actually I still prefer just to specify feature, I'd mark `youtube-dlc` as deprecated but don't know how.

vilgotf:
Note that build scripts for env vars won't be necessary as of 1.56 ([source](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#configurable-env)).

FelixMcFelix:
I discussed this with Mawa on Discord: I think the best fix either now or later is to have a `ytdl_like` function which allows the user to also specify an executable. This is probably preferable to a feature explosion with all these forks...

-- Kyle

Miezhiko:
@FelixMcFelix mawa is me, I didn't understood idea with `ytdl_like` function, but env variable with  what @vilgotf posted in last message could be fine I think, however still possible known forks should be documented and yet I think dlc should be marked as deprecated and maybe for only one or two forks current way is okay.

FelixMcFelix (approved):
Apologies it's been so long to get to this. This is simple enough for now; I'm not fond of this design in the long run, so I'd like to change/deprecate it on `next` before long to something more generic.

PR #102, opened 2021-10-19
Gateway: Add generics to `Call` methods.

### LINKED ISSUE #94: Why not generics?
Why [Songbird::get](https://docs.rs/songbird/0.2.0/songbird/struct.Songbird.html#method.get) have generics (`<G: Into<GuildId>>`), but [Songbird::get_or_insert](https://docs.rs/songbird/0.2.0/songbird/struct.Songbird.html#method.get_or_insert) doesn't have generics?
And, why [Call](https://docs.rs/songbird/0.2.0/songbird/struct.Call.html) methods doesn't have generics?

FelixMcFelix:
Adds generics to any `Id` types on `Call`. Also includes the overlooked `Songbird::get_or_insert`.

Closes #94. Tested using `cargo make ready`.

### DIFF
--- a/src/handler.rs
+++ b/src/handler.rs
@@ -10,6 +10,7 @@ use crate::{
 };
 use flume::Sender;
 use serde_json::json;
+use std::fmt::Debug;
 use tracing::instrument;
 
 #[cfg(feature = "driver-core")]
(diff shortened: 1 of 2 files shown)


vilgotf:
Is `Debug` really necessary? They get turned into `GuildId` (which impl `Default`)

arqunis:
The `Debug` bound is necessary for the [`#[instrument]`](https://docs.rs/tracing/0.1.29/tracing/attr.instrument.html) macro.

PR #105, opened 2021-12-06
Support serenity simd_json

vicky5124:
modified:   src/shards.rs


src/shards.rs
```diff
@@ -5,10 +5,13 @@ use crate::error::{JoinError, JoinResult};
 use futures::channel::mpsc::{TrySendError, UnboundedSender as Sender};
 #[cfg(feature = "serenity")]
 use parking_lot::{lock_api::RwLockWriteGuard, Mutex as PMutex, RwLock as PRwLock};
+#[cfg(not(feature = "serenity"))]
 use serde_json::Value;
 #[cfg(feature = "serenity")]
```
FelixMcFelix: Could you please combine your new import into this one here?

FelixMcFelix (changes requested):
Thanks for the PR, just one minor nit regarding import formatting. I had a quick look at the repo and it seems that voice-model is untouched, so I think this is the extent of everything you need to change. :)

vicky5124:
sorry! should have marked this as a draft, as there's still way more files to modify, and things that need to be fixed on serenity first before this can work and be merged.
See https://github.com/serenity-rs/serenity/pull/1613

About the import order, cargo-fmt made it that way.

src/shards.rs
vicky5124: This is possibly a side effect of cargo-fmt

.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustflags = ["-C", "target-cpu=haswell"]
```
FelixMcFelix: Why not `native`?

Can this also be moved to CI such that it's only invoked for the simd-json build task? I'm unsure if the spurious MacOS SIGILL originates from this.

src/input/dca.rs
```diff
@@ -48,6 +48,12 @@ async fn _dca(path: &OsStr) -> Result<Input, DcaError> {
 
     let reader = json_reader.into_inner().into_std().await;
 
+    #[cfg(feature = "serenity")]
+    let metadata: Metadata =
+        serenity::json::prelude::from_slice::<DcaMetadata>(raw_json.as_mut_slice())
+            .map_err(DcaError::InvalidMetadata)?
+            .into();
+    #[cfg(not(feature = "serenity"))]
```
FelixMcFelix: General feedback for all other driver/input uses: wouldn't it make more sense to use the simd-json library ourselves if that feature is selected? That way we can apply this to twilight too, and use `serenity::json::*` only where we interface with serenity in the gateway portions. Twilight v0.8 now exposes their own JSON type via `twilight_gateway::shard::raw_message::Message` from what I can see.

FelixMcFelix (changes requested):
Currently relies on serenity's simdjson support being fixed before I can end-to-end test this.

Otherwise compiles fine, comments are inline with respect to all driver-related JSON parsing.

.cargo/config.toml
vicky5124: No, unless you do CI with Arm, this target-cpu will work just fine.
Haswell was chosen over native as I've seen a few people mention it's less likely for it to fail CI due to "Illegal Instruction"; MacOS failed because there's still a chance it can fail from that, serenity has had this issue for quite a while now, and it's on Actions for it to be fixed.

Moving this to only the SIMD check would make it more likely to pass, and I'll do it when I have time for it, but it's hard to test CI when it's blocked behind it being accepted due to this being my first PR.

src/input/dca.rs
vicky5124: I somewhat planned for this, but I have 0 idea how twilight manages SIMD; I will not make a PR of something I'm completely unfamiliar with. If someone else wants to implement the SIMD feature for twilight, go for it.

vicky5124:
Serenity has fixed it's SIMD issues, this PR is ready to be tested.

FelixMcFelix:
Tried again there, still getting plenty of messages like below after a `cargo update`:

```
Jan 08 17:07:25.284  WARN run:recv_event:handle_event{event=Err(SimdJson(Error { index: 0, character: '💩', error: Serde("event GuildCreate: Serde(\"Serde(\\\"Serde(\\\\\\\"invalid type: string \\\\\\\\\\\\\\\"925690819955023872\\\\\\\\\\\\\\\", expected tuple struct MessageId\\\\\\\") at character 0 ('💩')\\\") at character 0 ('💩')\") at character 0 ('💩')") }))}: serenity::gateway::shard: [Shard [0, 1]] Unhandled error: SimdJson(Error { index: 0, character: '�'', error: Serde("event GuildCreate: Serde(\"Serde(\\\"Serde(\\\\\\\"invalid type: string \\\\\\\\\\\\\\\"925690819955023872\\\\\\\\\\\\\\\", expected tuple struct MessageId\\\\\\\") at character 0 ('💩')\\\") at character 0 ('💩')\") at character 0 ('�'')") })
```

nickelc:
> Tried again there, still getting plenty of messages like below after a `cargo update`:
> 
> ```
> Jan 08 17:07:25.284  WARN run:recv_event:handle_event{event=Err(SimdJson(Error { index: 0, character: '💩', error: Serde("event GuildCreate: Serde(\"Serde(\\\"Serde(\\\\\\\"invalid type: string \\\\\\\\\\\\\\\"925690819955023872\\\\\\\\\\\\\\\", expected tuple struct MessageId\\\\\\\") at character 0 ('💩')\\\") at character 0 ('💩')\") at character 0 ('💩')") }))}: serenity::gateway::shard: [Shard [0, 1]] Unhandled error: SimdJson(Error { index: 0, character: '�'', error: Serde("event GuildCreate: Serde(\"Serde(\\\"Serde(\\\\\\\"invalid type: string \\\\\\\\\\\\\\\"925690819955023872\\\\\\\\\\\\\\\", expected tuple struct MessageId\\\\\\\") at character 0 ('💩')\\\") at character 0 ('💩')\") at character 0 ('�'')") })
> ```

I think that was an issue in `simd-json` and should be fixed now.

FelixMcFelix:
I'll manually fixup and test after #89 is merged (i.e., after a cleanup pass). Apologies that this has been left for quite so long.

FelixMcFelix:
Okay, I've fixed this up to work without serenity now (and I believe JSON decode in DCA format and YoutubeDL handling should be accelerated too). This might take some fiddling to make sure the CI plays nice -- I've tried to make sure that RUSTFLAGS are only ever set on the Linux/simd target (since MacOS breaks routinely).

PR #108, opened 2022-01-11
allow users to make their own childcontainers / readers

asg0451:


FelixMcFelix (approved):
Thanks, looks good to me. Users should *technically* already be able to do this through `From<Vec<Child>>`, but this is more obvious. In #89 I've changed the inner field to be `pub`, but this suffices for now since that's breaking. :)

asg0451:
Ah I didn't see that. Yeah this is more obvious agreed. Thanks!

PR #111, opened 2022-01-26
Fix typo in CHANGELOG.md

reiyw:
This is a very minor issue, but I found that the month and day were swapped, so I'm sending the fix.


FelixMcFelix (approved):
Good catch, thanks!

PR #116, opened 2022-03-07
feat: return track handle when adding an `Input` to the queue

vaporoxx:
This modifies `Driver::enqueue_source` and `TrackQueue::add_source` to return the created track handle. This can be useful if you want to e.g. access the metadata after adding it and makes it more consistent with `Driver::play_source`, which returns the handle as well.


FelixMcFelix (approved):
Thanks, this is a good ergonomic fix.

PR #118, opened 2022-03-30
Remove tokio 0.2 support

GnomedDev:
> ```
> # Tokio 0.2 Compatibility features
> # These should probably be dropped around the same time as serenity drop them.```
Well... https://github.com/serenity-rs/serenity/pull/1780 got merged so it's time to remove tokio 0.2 support on songbird

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,9 +56,6 @@ jobs:
           - name: gateway only
             features: serenity-rustls
             dont-test: true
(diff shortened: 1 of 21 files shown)


FelixMcFelix (changes requested):
Thanks for taking this on, looks good. Only necessary change is to modify the CI to remove the legacy tokio task, and this should be golden.

FelixMcFelix (approved):
Great, LGTM. Thanks!

PR #121, opened 2022-04-19
serenity 11 support, drop legacy tokio support

Miezhiko:


FelixMcFelix:
These are already covered on `next` by dcb33eba101dd4aefc312b5ce495268da085f231 and #118.

PR #124, opened 2022-04-22
Remove spin_sleep in `Mixer::march_deadline`

GnomedDev:
There is no point to be that precise when network latency will ruin it, plus it was using 28% of CPU cycles on my bot.


FelixMcFelix:
Examples are broken for an unrelated reason, otherwise looks good! Thanks.

PR #142, opened 2022-07-29
Fix typo in docs for VoiceData

tazz4843:
The docs stated that decoded audio was mono, where it was actually stereo. This PR fixes that.


FelixMcFelix (approved):
Simple enough, and good catch on that oversight. Thanks!

PR #147, opened 2022-08-19
Update twilight to 0.13

Erk-:
Update twilight to 0.13

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -359,7 +359,8 @@ impl Songbird {
     pub async fn process(&self, event: &TwilightEvent) {
         match event {
             TwilightEvent::VoiceServerUpdate(v) => {
-                let call = v.guild_id.map(GuildId::from).and_then(|id| self.get(id));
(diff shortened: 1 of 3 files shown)


src/manager.rs
```diff
@@ -359,7 +359,8 @@ impl Songbird {
     pub async fn process(&self, event: &TwilightEvent) {
         match event {
             TwilightEvent::VoiceServerUpdate(v) => {
-                let call = v.guild_id.map(GuildId::from).and_then(|id| self.get(id));
+                let id = GuildId::from(v.guild_id);
+                let call =  self.get(id);
```
FelixMcFelix: I think this is setting off the linter!

```suggestion
                let call = self.get(id);
```

FelixMcFelix (changes requested):
Simple enough, thanks! I think there's one nit causing rustfmt to complain I've pointed out, otherwise should be fine.

FelixMcFelix (approved):

PR #151, opened 2022-09-18
Avoid spawning a disposal thread per driver

GnomedDev:
*This is part of my work to reduce the thread count of Songbird, to majorly improve performance and reduce load on the scheduler.*

Adds a new field to Config, `disposer`, an `Option<Sender<DisposalMessage>>` responsible for dropping the `DisposalMessage` on a separate thread.

If this is not set, and the `Config` is passed into `manager::Songbird`, a thread is spawned for this purpose (which previously was spawned per driver). 
If this is not set, and the `Config` is passed directly into `Driver` or `Call`, a thread is spawned locally, which is the current behavior as there is no where to store the `Sender`.

This `disposer` is then used in `Driver` as previously, to run possibly blocking destructors (which should only block the disposal thread). I cannot see this disposal thread getting overloaded, but if it is the `DisposalMessage`s will simply be queued in the flume channel until it can be dropped.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,15 +2,13 @@
 use crate::driver::DecodeMode;
 #[cfg(feature = "driver")]
 use crate::{
-    driver::{retry::Retry, tasks::disposal, tasks::message::DisposalMessage, CryptoMode, MixMode},
+    driver::{retry::Retry, tasks::disposal::DisposalThread, CryptoMode, MixMode},
     input::codecs::*,
 };
 
 #[cfg(test)]
 use crate::driver::test_config::*;
 
-#[cfg(feature = "driver")]
-use flume::Sender;
 #[cfg(feature = "driver")]
 use symphonia::core::{codecs::CodecRegistry, probe::Probe};
 
@@ -154,7 +152,7 @@ pub struct Config {
     /// Note: When using [`Songbird`] this is overwritten automatically by it's disposal thread.
     ///
     /// [`Songbird`]: crate::Songbird
-    pub disposer: Option<Sender<DisposalMessage>>,
+    pub disposer: Option<DisposalThread>,
 
     // Test only attributes
     #[cfg(feature = "driver")]
@@ -280,7 +278,7 @@ impl Config {
 
     /// Sets this `Config`'s channel for sending disposal messages.
     #[must_use]
-    pub fn disposer(mut self, disposer: Sender<DisposalMessage>) -> Self {
+    pub fn disposer(mut self, disposer: DisposalThread) -> Self {
         self.disposer = Some(disposer);
         self
     }
@@ -291,7 +289,7 @@ impl Config {
         if self.disposer.is_some() {
             self
         } else {
-            self.disposer(disposal::run())
+            self.disposer(DisposalThread::run())
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -143,6 +145,16 @@ pub struct Config {
     ///
     /// [`PROBE`]: static@PROBE
     pub format_registry: &'static Probe,
+    #[cfg(feature = "driver")]
+    /// The Sender for a channel that will run the destructor of possibly blocking values.
+    ///
+    /// If not set, a thread will be spawned to perform this, but it is recommended to create
+    /// a long running thread instead of relying on a per-driver thread.
+    ///
+    /// Note: When using [`Songbird`] this is overwritten automatically by it's disposal thread.
```
FelixMcFelix:
```suggestion
    /// Note: When using [`Songbird`] this is overwritten automatically by its disposal thread.
```

src/driver/tasks/mixer/mod.rs
```diff
@@ -126,14 +126,11 @@ impl Mixer {
...
 
-        // Create an object disposal thread here.
-        let (disposer, disposal_rx) = flume::unbounded();
-        std::thread::spawn(move || disposal::runner(disposal_rx));
-
         let thread_pool = BlockyTaskPool::new(async_handle);
 
         let symph_layout = config.mix_mode.symph_layout();
 
+        let disposer = config.disposer.clone().unwrap_or_else(disposal::run);
```
FelixMcFelix: Good, this catches an edge case I had in mind where a `None` might be passed in.

src/config.rs
```diff
@@ -143,6 +145,16 @@ pub struct Config {
...
+    ///
+    /// [`Songbird`]: crate::Songbird
+    pub disposer: Option<Sender<DisposalMessage>>,
```
FelixMcFelix: `Config` is the right place for this to be set, but I'm unhappy with the type `Sender` being exposed since that now makes flume upgrades into breaking changes.

Is the right choice maybe to make an opaque newtype to handle this?
```rust
pub struct DisposalThread(Sender<DisposalMessage>);
```
I guess this has a really obvious `Default` impl which you can lean on elsewhere, and still allows for users to set up whatever number and allocation of disposal threads they want.

You might be able to define a `DisposalThread::with_fn` as well to take a closure/fn, but I don't think that's a good idea as users shouldn't be able to access `InnerTracks` etc. anyway.

FelixMcFelix (changes requested):
This is a good idea and I like how it's handled mostly, I only have issues with exposing `Sender` and similar types to users.

src/config.rs
GnomedDev: Do you mean by the default impl to spawn the thread in that? If so, that solves the issue I had in mind of the disposal sender being used for stuff other than disposing....

FelixMcFelix: Exactly, so something like:

```rust
impl Default for DisposalThread {
    pub fn default() -> Self {
        Self(disposal::run)
    }
}
```

You'd likely need a `pub(crate)` on the inner `Sender` field, after some reflection...

GnomedDev: Done! Didn't need to pub(crate) it and cleaned up some stuff in the process

FelixMcFelix (approved):
Thanks very much, looks good to me!

PR #152, opened 2022-10-10
Fix wrong docstring for Track::volume

btoschek:
This is a very minor issue, but I found a wrong docstring for Track::volume. Sending the fix.


FelixMcFelix:
Thank you, and good catch!

PR #154, opened 2022-11-29
Fixed position reporting bug

JamesDSource:
Without this line resetting the position here, two things happen

- When seeking backwards in a track, the reported position will just add the time you seek to itself
- When a track loops, it does not reset it's position to 0 and will continue to overflow from it's end time


FelixMcFelix (approved):
Sorry for the delay, this makes sense. Thanks!

PR #166, opened 2023-03-21
Add `iter` to `Songbird`

### LINKED ISSUE #165: There should be a way to get all `Arc<Mutex<Call>>`s
**Songbird version:** next

**Description:**

I'm currently implementing a procedure for saving call information (including tracks from the builtin queue) to disk, but it looks like the `Songbird` type does not provide access to iter over all calls? Should this feature be added or am I approaching it in a wrong way?

fee1-dead:
Closes #165.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -185,6 +185,16 @@ impl Songbird {
         })
     }
 
+    /// Creates an iterator for all [`Call`]s currently managed.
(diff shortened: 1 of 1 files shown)


FelixMcFelix (approved):
Looks good to me. Cloning & copying KV pairs should avoid any Dashmap reference locking issues, so this is a clean enough solution. Thanks!

PR #168, opened 2023-03-22
Input: Pass `--no-playlist` for `YoutubeDl`

fee1-dead:
If a link such as [this](https://www.youtube.com/watch?v=ygY2qObZv24&list=RDygY2qObZv24) is passed to `YoutubeDl` without the option, it would cause a deadlock in my bot.

There were many videos where it produced a `Silent` packet first instead of mixed or passthrough. The URL I added was one that produced a passthrough packet so I used that. Please let me know if this is wrong.


FelixMcFelix:
I think this is a sensible filter to apply when URL scraping, thanks for finding the issue.

PR #170, opened 2023-04-08
fix(ws): Songbird would fail if it could not deserialize ws payload.

Erk-:
Discord started sending a new event on the voice websocket with opcode 18:
```json
{"op":18,"d":{"user_id":"...","flags":0}}
```
This caused songbird to crash and disconnect.

This patch makes it so that failing to deserialize a event is not a fatal error.

### DIFF
--- a/src/ws.rs
+++ b/src/ws.rs
@@ -10,7 +10,7 @@ use async_tungstenite::{
 use futures::{SinkExt, StreamExt, TryStreamExt};
 use serde_json::Error as JsonError;
 use tokio::time::{timeout, Duration};
-use tracing::instrument;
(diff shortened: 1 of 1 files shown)


Skarlett:
updated my project with pr-branch, appears to work great.

https://github.com/Skarlett/coggie-bot/commit/1986feae8ca50e2a2924ccdd04e22e2ca9f243cb

FelixMcFelix (approved):
This is effectively moving the check from the main ws runner loop to the WS stream itself, which I think is a better solution. I do need to clean up the main WS loop to remove the check there but pushing this through for now.

PR #171, opened 2023-04-08
Twilight 0.15 support

Erk-:
This patch changes around quite a few things.
The main entry point for twilight besides `process` will now be the
`TwilightMap` which consists of command senders for each shard.

This simplifies parts of the code as there is not any difference
between shards and clusters anymore.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,8 +44,8 @@ tokio-tungstenite = { optional = true, version = "0.18" }
 tokio-util = { features = ["io"], optional = true, version = "0.7" }
 tracing = { version = "0.1", features = ["log"] }
 tracing-futures = "0.2"
-twilight-gateway = { default-features = false, optional = true, version = "0.14.0" }
-twilight-model = { default-features = false, optional = true, version = "0.14.0" }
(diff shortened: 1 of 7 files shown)


src/ws.rs
```diff
@@ -96,8 +96,19 @@ pub(crate) fn convert_ws_message(message: Option<Message>) -> Result<Option<Even
         // simd-json::serde::from_str may leave an &mut str in a non-UTF state on failure.
         // The below is safe as we have taken ownership of the inner `String`, and don't
         // access it as a `str`/`String` or return it if failure occurs.
-        Some(Message::Text(mut payload)) =>
-            unsafe { crate::json::from_str(payload.as_mut_str()) }.map(Some)?,
+        Some(Message::Text(mut payload)) => {
```
FelixMcFelix: You've accidentally included #170 here, but since the merge will already be really messy we should leave it for now to prevent complaints over the weekend.

FelixMcFelix (changes requested):
Looks good to me, just missing a `cargo +nightly fmt --all` according to CI.

FelixMcFelix (approved):
Great, good work. Thanks!

(CI failures are due to divergence from Serenity, unrelated to this as far as I know)

PR #174, opened 2023-04-10
Move error handling

Erk-:
Move the place where we look away from the json error to be higher up as discussed on Discord


FelixMcFelix (approved):
Apologies for pushing direct to your branch, please let me know if you're happy with this approach as well.

Basically, the reconnection pathway after an error will *also* be hit by the same bug, so all checks are moved into `convert_ws_message`. Given that Discord will now apparently send other WS events during the handshake, I've also demoted all non-{Hello, Ready, Resumed} messages during handshakes to logs rather than an explicit error,

PR #175, opened 2023-04-23
Chore(workflows): Update checkout action from v2 to v3

jontze:
This change was done as the v2 version still used the node v12 runtime. This runtime is deprecated by github. There are no other breaking changes done by the action. With the update to v3 now node v16 is used as runtime for the action.

You can read more about the changes from v2 to v3 here: https://github.com/actions/checkout/releases/tag/v3.0.0

Thanks for the cool project. I really enjoy serenity and songbird! :heart: 

There are a few other deprecated actions in the workflows. I plan to update these as well in other PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
 
     steps:
     - name: Checkout sources
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


FelixMcFelix (approved):
Thanks! I'm aware the CI pipelines are stagnating a bit, so anything helps out a lot.

PR #177, opened 2023-04-24
Chore(ci): Update rust, cargo and cache actions

jontze:
This PR moves the caching work to the [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache) action and moved away from the deprecated actions of the [actions-rs](https://github.com/actions-rs) github organization.

The actual building, testing and linting stays untouched.

I removed one job from the test matrix, as the `legacy tokio` job doesn't seem to be necessary any more.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,6 @@ jobs:
           - Windows
           - driver only
           - gateway only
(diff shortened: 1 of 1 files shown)


FelixMcFelix (approved):
Thanks again, this is far tidier! LGTM.

Issue #184, opened 2023-05-19
Intermittent Symphonia Probe Failures causing Playback issues on `next` branch


maxall41:
**Songbird version:** `next`

**Rust version (`rustc -V`):** 1.69

**Serenity/Twilight version:** `next` branch serenity

**Description:**
I'm getting intermittent symphonia `reached probe limit` errors on the `next` branch with about a 1/50 to a 1/5 chance of happening. It's a bit hard to tell because of the low sample size. It does seem to increase with more load  (more concurrent streams). But its not a large load about 5 concurrent streams with 11% CPU Usage and only 30MB Memory usage. And i'm not entirely sure that is the case. Here is the log section when it happens:
```
[2023-05-19T00:17:08Z INFO songbird::driver::tasks::events ] Global event added.
[2023-05-19T00:17:10Z INFO songbird::driver::tasks::events ] Event state for track 1 added
[2023-05-19T00:17:10Z INFO songbird::events::store ] Firing Preparing for [0]
[2023-05-19T00:17:11Z ERROR symphonia_core::probe ] reached probe limit of 1048576 bytes.
[2023-05-19T00:17:11Z INFO songbird::events::store ] Firing Error for [0]
[2023-05-19T00:17:11Z ERROR hearth::worker::errors ] Track 23810d79-74a4-4a0e-bf8a-ed795d2db073 encountered an error: Errored(Parse(Unsupported("core (probe): no suitable format reader found")))
```
This happens when using the HttpRequest source. Though it may happen with others as I haven't tested any others.

**Steps to reproduce:**
1. Copy basic example
2. Create concurrent streams
3. Try stopping and playing streams until error occurs.

**Environment**
Arch: x86
OS: Ubuntu Linux
System: 4 Core 8GB

FelixMcFelix:
Whenever I saw something like this in testing, it tended to be because I'd been served a page body or other HTTP response instead of valid audio data. What website are you making requests to? What codecs do you have enabled?

maxall41:
The funny thing is I'm using the exact same file served from the exact same URL between play requests. And it seems to happen intermittently. And I'm not changing anything between requests. The file I'm using is: https://www.ee.columbia.edu/~dpwe/sounds/music/africa-toto.wav. And the codecs I have enabled are:
```toml
symphonia = { version = "0.5.2", features = ['pcm','mp3','wav','isomp4','aac','alac'] }
```
I also made sure I'm not getting rate limited or anything

[...]

PR #190, opened 2023-06-01
HTTP Status Code Checks

maxall41:
Checks if the status code returned from an http request in the HttpRequest source is between 200 - 299 (OK). Otherwise returns error. Resolves issue #184

### DIFF
--- a/src/input/error.rs
+++ b/src/input/error.rs
@@ -18,8 +18,6 @@ pub enum AudioStreamError {
     Fail(Box<dyn Error + Send + Sync>),
     /// The operation was not supported, and will never succeed.
     Unsupported,
-    /// The Http request failed. If you get this you may be being rate limited
-    HttpRequestFailed(String),
 }
 
 impl Display for AudioStreamError {
@@ -29,9 +27,6 @@ impl Display for AudioStreamError {
             Self::RetryIn(t) => f.write_fmt(format_args!("retry in {:.2}s", t.as_secs_f32())),
             Self::Fail(why) => f.write_fmt(format_args!("{why}")),
             Self::Unsupported => f.write_str("operation was not supported"),
-            Self::HttpRequestFailed(code) => f.write_fmt(format_args!(
-                "Got HTTP Response Code: {code}. You may be getting rate limited."
-            )),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/input/error.rs
```diff
@@ -18,6 +18,8 @@ pub enum AudioStreamError {
     Fail(Box<dyn Error + Send + Sync>),
     /// The operation was not supported, and will never succeed.
     Unsupported,
+    /// The Http request failed. If you get this you may be being rate limited
```
FelixMcFelix: This `Error` enum isn't really designed to be extended. A non-2xx request should map to an `AudioStreamError::Fail(...)` (I think a 503 with a Retry-after header is the exception, but that's out of scope). The reason we do this is so that the internals have a limited number of error cases to handle, many of which could be user code (source creation, or occur during decode).

Could you please make use of the `Fail` error variant rather than creating a new one?

src/input/error.rs
```diff
@@ -27,6 +29,9 @@ impl Display for AudioStreamError {
             Self::RetryIn(t) => f.write_fmt(format_args!("retry in {:.2}s", t.as_secs_f32())),
             Self::Fail(why) => f.write_fmt(format_args!("{why}")),
             Self::Unsupported => f.write_str("operation was not supported"),
+            Self::HttpRequestFailed(code) => f.write_fmt(format_args!(
+                "Got HTTP Response Code: {code}. You may be getting rate limited."
```
FelixMcFelix: In case you want to modify this setup into a file-local error (e.g., an `Error` type in `http.rs` that isn't exported), the convention is that `Display` strings for error types are written in lower case. See [the trait documentation](https://doc.rust-lang.org/std/error/trait.Error.html) -- "Error messages are typically concise lowercase sentences without trailing punctuation".

src/input/sources/http.rs
```diff
@@ -82,6 +82,12 @@ impl HttpRequest {
             .await
             .map_err(|e| AudioStreamError::Fail(Box::new(e)))?;
 
+        if !resp.status().is_success() {
+            return Err(AudioStreamError::HttpRequestFailed(
+                resp.status().to_string(),
```
FelixMcFelix: As above, please convert into an `AudioStreamError::Fail`. You can either place a string in here directly (as it should be coerceable to `Box<dyn Error>`) or you can perhaps setup a local `Error` struct which stores the status code and/or body for use in a `Display` impl.

FelixMcFelix (changes requested):
Thanks for your work, I've left some comments on how to make use of the existing error types; I apologise that I've not really documented the "developer intent" around how they should be treated.

FelixMcFelix (approved):
LGTM, thanks!

PR #198, opened 2023-07-29
Replace `xsalsa20poly1305` with `crypto_secretbox`

Sebbl0508:
`xsalsa20poly1305` is listed as deprecated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,11 @@ version = "0.3.0-rc.0"
 optional = true
 version = "1"
 
+[dependencies.crypto_secretbox]
+optional = true
+version = "0.1.1"
+features = ["std"]
+
 [dependencies.dashmap]
 optional = true
 version = "5"
@@ -109,11 +114,6 @@ optional = true
 version = "0.8"
 features = ["v4"]
 
-[dependencies.crypto_secretbox]
-optional = true
-version = "0.1.1"
-features = ["std"]
-
 [dev-dependencies]
 criterion = "0.3"
 utils = { path = "utils" }
@@ -152,6 +152,7 @@ driver-core = [
     "async-trait",
     "audiopus",
     "byteorder",
+    "crypto_secretbox",
     "discortp",
     "flume",
     "parking_lot",
@@ -161,7 +162,6 @@ driver-core = [
     "typemap_rev",
     "url",
     "uuid",
-    "crypto_secretbox",
 ]
 rustls = ["async-tungstenite/tokio-rustls-webpki-roots", "rustls-marker"]
 native = ["async-tungstenite/tokio-native-tls", "native-marker"]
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -10,7 +10,7 @@ license = "ISC"
 name = "songbird"
 readme = "README.md"
 repository = "https://github.com/serenity-rs/songbird.git"
-version = "0.3.2"
+version = "0.3.3"
```
FelixMcFelix: Please don't change this: we'll handle this separately when we make a release at some later time.

Cargo.toml
```diff
@@ -109,9 +109,9 @@ optional = true
 version = "0.8"
 features = ["v4"]
 
-[dependencies.xsalsa20poly1305]
+[dependencies.crypto_secretbox]
```
FelixMcFelix: Can you move the `[dependencies.crypto_secretbox]` block to its correct alphabetical place? I.e., between byteorder and dashmap.

Cargo.toml
```diff
@@ -161,7 +161,7 @@ driver-core = [
     "typemap_rev",
     "url",
     "uuid",
-    "xsalsa20poly1305",
+    "crypto_secretbox",
```
FelixMcFelix: Ditto on alphabetical order.

src/driver/crypto.rs
```diff
@@ -241,8 +244,10 @@ impl CryptoState {
 #[cfg(test)]
 mod test {
     use super::*;
+    use crypto_secretbox::{KeyInit, SecretBox};
     use discortp::rtp::MutableRtpPacket;
-    use xsalsa20poly1305::{aead::NewAead, KEY_SIZE, TAG_SIZE};
+    pub const KEY_SIZE: usize = SecretBox::<()>::KEY_SIZE;
+    pub const NONCE_SIZE: usize = SecretBox::<()>::KEY_SIZE;
```
FelixMcFelix: I don't think this line is necessary, and is also incorrect (`NONCE` != `KEY`).

src/driver/tasks/mixer.rs
```diff
@@ -20,7 +21,8 @@ use rand::random;
 use std::{convert::TryInto, time::Instant};
 use tokio::runtime::Handle;
 use tracing::{debug, error, instrument};
-use xsalsa20poly1305::TAG_SIZE;
+
+const TAG_SIZE: usize = SecretBox::<()>::TAG_SIZE;
```
FelixMcFelix: Why not `use crate::driver::crypto::TAG_SIZE`?

FelixMcFelix (changes requested):
Thanks, I wasn't aware that the library had been deprecated. There are some minor changes I'd like to ask for, otherwise I can see it's mostly a drop-in replacement.

Sebbl0508:
I've adjusted the code. hope it looks good now

FelixMcFelix (approved):
Great, looks good to me! Thank you. :)

PR #199, opened 2023-10-08
Fix compiling with latest serenity

GnomedDev:
Serenity ID internals are now hidden, so we have to use get -> new -> unwrap.


FelixMcFelix (approved):

FelixMcFelix:
Thanks!

PR #204, opened 2023-11-08
Fix clippy pedantic warnings

GnomedDev:


FelixMcFelix (approved):
Thanks for running through this, looks good. Judging from how many of these are in test-cfg'd code, we might need to add `--all-targets` to the standard clippy runs + CI in future.

PR #207, opened 2023-11-28
 Replace OnceCell with std::sync::OnceLock 

GnomedDev:
OnceCell got merged into the rust stdlib a while ago, so let's use it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,7 @@ Thanks to the following for their contributions:
 
 ### Fixed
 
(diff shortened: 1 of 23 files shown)


FelixMcFelix (changes requested):
Looks good, but there are still apparently a few doc links to `CODEC_REGISTRY` and `PROBE` needing cleaned up. Should be good to merge after you catch those.

FelixMcFelix (approved):
Thanks!

PR #210, opened 2023-12-04
Add ytdl search

cycle-five:
Adds to YoutubeDl the ability to just return the search results as text.

### DIFF
--- a/src/input/compose.rs
+++ b/src/input/compose.rs
@@ -37,4 +37,11 @@ pub trait Compose: Send {
     async fn aux_metadata(&mut self) -> Result<AuxMetadata, AudioStreamError> {
         Err(AudioStreamError::Unsupported)
     }
+
+    /// Uses a query to search for a source and return the results.
+    ///
+    /// This method allows for searching for a source and being able to select from the results.
+    async fn search(&mut self) -> Result<Vec<String>, AudioStreamError> {
+        Err(AudioStreamError::Unsupported)
+    }
 }
--- a/src/input/compose.rs
+++ b/src/input/compose.rs
@@ -37,11 +37,4 @@ pub trait Compose: Send {
     async fn aux_metadata(&mut self) -> Result<AuxMetadata, AudioStreamError> {
         Err(AudioStreamError::Unsupported)
     }
-
-    /// Uses a query to search for a source and return the results.
-    ///
-    /// This method allows for searching for a source and being able to select from the results.
-    async fn search(&mut self) -> Result<Vec<String>, AudioStreamError> {
-        Err(AudioStreamError::Unsupported)
-    }
 }
--- a/src/input/sources/ytdl.rs
+++ b/src/input/sources/ytdl.rs
@@ -12,7 +12,7 @@ use reqwest::{
     header::{HeaderMap, HeaderName, HeaderValue},
     Client,
 };
-use std::{error::Error, io::ErrorKind};
+use std::{error::Error, io::ErrorKind, time::Duration};
 use symphonia_core::io::MediaSource;
 use tokio::process::Command;
 
@@ -56,17 +56,22 @@ impl YoutubeDl {
         }
     }
 
-    /// Creates a lazy(?) request to search youtube for a videos matching `query`.
+    /// Creates a request to search youtube for an optionally specified number of videos matching `query`.
     ///
     /// [`new`]: Self::new
     #[must_use]
-    pub fn new_yt_search(client: Client, query: String) -> Self {
-        Self::new_ytdl_like(YOUTUBE_DL_COMMAND, client, format!("ytsearch5:{}", query))
+    pub fn new_yt_search(client: Client, query: String, results: Option<u32>) -> Self {
+        let results = results.unwrap_or(5);
+        Self::new_ytdl_like(
+            YOUTUBE_DL_COMMAND,
+            client,
+            format!("ytsearch{}:{}", results, query),
+        )
     }
 
     /// Does a search query for the given url, returning a list of possible matches
     /// which are youtube urls.
-    pub async fn search_query(&mut self) -> Result<Vec<String>, AudioStreamError> {
+    pub async fn search_query(&mut self) -> Result<Vec<AuxMetadata>, AudioStreamError> {
         let search_str = if self.url.starts_with("ytsearch") {
             self.url.clone()
         } else {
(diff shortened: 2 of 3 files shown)


cycle-five:
What are your linter settings? Mine seem to be conflicting.

GnomedDev:
You need to run `cargo +nightly fmt`.

src/input/compose.rs
```diff
@@ -37,4 +37,11 @@ pub trait Compose: Send {
     async fn aux_metadata(&mut self) -> Result<AuxMetadata, AudioStreamError> {
         Err(AudioStreamError::Unsupported)
     }
+
+    /// Uses a query to search for a source and return the results.
+    ///
+    /// This method allows for searching for a source and being able to select from the results.
+    async fn search(&mut self) -> Result<Vec<String>, AudioStreamError> {
```
FelixMcFelix: I don't think that this should be added to the `Compose` trait -- this is not common functionality to all or most Input sources.

src/input/sources/ytdl.rs
```diff
@@ -143,6 +189,10 @@ impl Compose for YoutubeDl {
             AudioStreamError::Fail(msg)
         })
     }
+
+    async fn search(&mut self) -> Result<Vec<String>, AudioStreamError> {
```
FelixMcFelix: See above re: `Compose::search`.

src/input/sources/ytdl.rs
```diff
@@ -56,6 +56,52 @@ impl YoutubeDl {
         }
     }
 
+    /// Creates a lazy(?) request to search youtube for a videos matching `query`.
```
FelixMcFelix:
```suggestion
    /// Creates a lazy request to search Youtube for the first video matching `query`.
```

src/input/sources/ytdl.rs
```diff
@@ -56,6 +56,52 @@ impl YoutubeDl {
...
+    ///
+    /// [`new`]: Self::new
+    #[must_use]
+    pub fn new_yt_search(client: Client, query: String) -> Self {
+        Self::new_ytdl_like(YOUTUBE_DL_COMMAND, client, format!("ytsearch5:{}", query))
```
FelixMcFelix: A) Why 5 specifically? Can this be made user-configurable? If you play this `Compose`, I'd expect it would just play the first so maybe `1` makes for a sensible default unless you want to use search query.

B) An `Input` created from this `Compose` will error out, as you will get 5-lines of newline delimited JSON from yt-dlp. You need to alter the JSON decode to account for this.

src/input/sources/ytdl.rs
```diff
@@ -56,6 +56,52 @@ impl YoutubeDl {
...
+                } else {
+                    Box::new(e)
+                })
+            })?;
+
+        let lines = output
+            .stdout
+            .split(|&b| b == b'\n')
+            .map(|line| format!("{}", String::from_utf8_lossy(line)))
+            .collect::<Vec<_>>();
```
FelixMcFelix: This isn't returning a list of valid URLs:
> `~play cloudkicker`

```rust
Ok([
"Cloudkicker - \"The Discovery\" [Full Album - Official - Correct Track Listing]",
"WVM603q6b1w",
"39:53",
"Best of Cloudkicker",
"_uxmzStcbss",
"3:27:04",
"Cloudkicker - Solitude [Album] (2020)",
"Ej1X5M0MAM8",
"48:32",
 // ...
])
```
This goes against the doc-comment of the function.

src/input/sources/ytdl.rs
```diff
@@ -56,6 +56,52 @@ impl YoutubeDl {
...
+            .args(ytdl_args)
+            .output()
+            .await
+            .map_err(|e| {
+                AudioStreamError::Fail(if e.kind() == ErrorKind::NotFound {
+                    format!("could not find executable '{}' on path", self.program).into()
+                } else {
+                    Box::new(e)
+                })
+            })?;
```
FelixMcFelix: While converting `query()` to support ndjson (`self.query_many(1).pop()`?), you could probably replace this function to:
* call an inner query function (`YoutubeDl::query_many(&mut self, n: usize)`?) to get a `Vec<AuxMetadata>`.
* You can then either extract `source_url` if you want a `Vec<String>` or just return all the metadata verbatim.

FelixMcFelix (dismissed):
I think that this feature is needed and useful, but there's work needing done to make it fit and function as expected if you're able to.

cycle-five:
I should have commented originally with the impetus for this feature. I am currently using it in my bot, albeit it is not fully implemented, however the entire purpose is to do a search and return options which the client can then choose from, before it plays the song.

src/input/sources/ytdl.rs
cycle-five: I will change the documentation, this is the intended output. I am open to a different format if it's preferred, but this is the information I am pulling purposefully for the feature in my bot I made this change for.

cycle-five:
> I think that this feature is needed and useful, but there's work needing done to make it fit and function as expected if you're able to.

I am more than happy to work on this to have it fit properly into songbird, it is absolutely needed for a feature I'm using in my bot, so as long as I can make that work, I'm happy to make whatever changes you feel are appropriate.

src/input/sources/ytdl.rs
FelixMcFelix: I think if you need all this, then returning `Vec<AuxMetadata>` is probably the right call for exposing all this in a structured way.

FelixMcFelix:
Thanks, I'm just keen that it works great for everyone. The main assumption with `Compose` is that you should be able to 'just play it', ideally. I.e., `new_yt_search` should produce an object that will play a search result, or produce a different struct entirely (which does not support `Compose`). I hope that makes sense?

src/input/compose.rs
cycle-five: removed

src/input/sources/ytdl.rs
cycle-five: removed

src/input/sources/ytdl.rs
cycle-five: fixed

src/input/sources/ytdl.rs
cycle-five: I like 5 and it seems reasonable number of search results for a discord bot result. I also made it configurable.

src/input/sources/ytdl.rs
cycle-five: done.

src/input/sources/ytdl.rs
cycle-five: I just returned it as AuxMetadata, lmk if the approach needs further change,

cycle-five:
I believe I've made all the requested changes and have got it working with these changes in my bot (which of course is the most important piece in this equation).

I feel that my approach to using the yt-dlp search feature and parsing that into the AuxMetadata is either exactly as it is intended, or completely wrong. 

Also for the future I believe yt-dlp may support other searches, and also are there plans to not need yt-dlp?

FelixMcFelix:
Thanks for making changes, I'll take a proper look in the next few days.

> I feel that my approach to using the yt-dlp search feature and parsing that into the AuxMetadata is either exactly as it is intended, or completely wrong.

It looks valid at a glance, thank you. If there are any changes then I'll either push to your branch or provide a patchset.

> Also for the future I believe yt-dlp may support other searches, 

We can support these as and when they come up.

> and also are there plans to not need yt-dlp?

You don't need yt-dlp for songbird, but practically speaking you need it to get any usable data from youtube given the way it works.

FelixMcFelix:
@cycle-five please have a look and see that you're okay with the changes I've made here:

* a search query is now playable as users would expect.
* search now calls into the same yt-dlp invocation and parsing logic as query.
* tests have been altered: relying on a specific test result's metadata is a bit flaky. Equally, we really want to make sure we're using sources which are creative-commons or similarly licensed.

cycle-five:
Awesome! I'm just going to pull this update and make sure my bot can integrate apropos.

cycle-five:
This works exactly as I need, so it has my OK.

PR #215, opened 2024-01-01
 Allow borrowed strings for YoutubeDl 

GnomedDev:
This changes `YoutubeDl` to have a lifetime, to allow users to provide `&str` references for the url/query. This can't currently be used for `Input`, where users have to provide `String` or `&'static str`, but is useful for users fetching metadata/searching.

This also makes `search` return `impl Iterator` to avoid the unnecessary `Vec` buffering possibly performed, which relies on `Either` as a primitive form of `enum_dispatch`, but it's already in the dep tree so it doesn't matter.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ crypto_secretbox = { optional = true, features = ["std"], version = "0.1" }
 dashmap = { optional = true, version = "5" }
 derivative = "2"
 discortp = { default-features = false, features = ["discord", "pnet", "rtp"], optional = true, version = "0.6" }
+either = "1.9.0"
 flume = { optional = true, version = "0.11" }
 futures = "0.3"
 nohash-hasher = { optional = true, version = "0.2.0" }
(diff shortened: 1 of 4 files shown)


FelixMcFelix (approved):
Thanks! This looks fine to me, apologies for taking so long to get around to this.

I was a bit thrown by program and query having the same lifetime -- I imagine in practice this covers queries pointing at `&'static str`s and `String`s, but maybe not building a `YoutubeDl` around a local non-static str to check metadata? Either way it's better than we have today, and I think it'd be a tad too noisy to add another lifetime param.

Feel free to merge when you're ready!

GnomedDev:
I think providing a 'static argument to program and a local `'a` to `YoutubeDl` will just shrink the lifetime to `'a`, but not too sure. I'm good for merge.

FelixMcFelix:
> I think providing a 'static argument to program and a local `'a` to `YoutubeDl` will just shrink the lifetime to `'a`,

Had a quick test on rust-playground, this is indeed the case. :)

PR #218, opened 2024-01-16
Fix serenity next breaking

GnomedDev:
Blocked on #219.

Serenity has made a switch from NonZero to NonMax for IDs, as it allows the cool niche opts without causing collisions with buggy discord code leading to this. We need to swap over to this.

Serenity has also reworked the state system, to remove typemap, so I fixed the examples.


src/tracks/handle.rs
```diff
@@ -17,22 +15,13 @@ use uuid::Uuid;
 /// to shared data not used by the driver.
 pub struct TrackHandle {
     inner: Arc<InnerHandle>,
+    data: Arc<dyn std::any::Any + Send + Sync + 'static>,
```
FelixMcFelix: I can see why you've done this, but moving this outside of InnerHandle to get an &mut won't work for reasons outlined below.

src/tracks/handle.rs
```diff
@@ -198,16 +189,32 @@ impl TrackHandle {
...
     }
 
-    /// Allows access to this track's attached [`TypeMap`].
+    /// Initialises or overwrites the given Track data.
     ///
-    /// [`TypeMap`]s allow additional, user-defined data shared by all handles
+    /// This type can be fetched again via [`Self::data`].
+    pub fn set_data<Data>(&mut self, data: Arc<Data>)
+    where
+        Data: Send + Sync + 'static,
```
FelixMcFelix: This won't set the data in any meaningful way. By the time a user can acquire a TrackHandle, it is a shared resource: one copy exists in the audio context, one copy in user code. A `set` here will not propagate the new data to any Handles returned alongside event handlers etc. The question is then how to get the chosen data into InnerHandle -- Arc::get_mut on a nested Arc won't work for the same reason.

I think it might be valid to put the `Arc<dyn Any>` into `Track` and source InnerHandle::data from there during handle creation? It can either be optioned or default to `Arc<() as _>`?This lets us continue avoiding locks like your new approach and keeps TrackHandle::clone bounded to a single Arc clone.

FelixMcFelix (changes requested):
I'm happy with the Id-focussed changes, and I think that serenity's new UserData functionality is a great improvement. Thanks for taking the time to port there.

The TrackHandle changes are a great idea in principle, but I think they need a little bit more iteration to work like they do today.

(Apologies for brevity/typos, reviewing from my phone under some time
constraints!)

src/tracks/handle.rs
GnomedDev: Ah, okay, I was somewhat unaware how this API was used. I'll have a look at moving it to Track.

GnomedDev: Moved to #219.

FelixMcFelix (approved):
Great, thanks for splitting out the other changes.

PR #219, opened 2024-01-17
Remove typemap from songbird

GnomedDev:
Split off from #218, removes typemap from songbird by using a dyn Any instead. This is already the approach taken for user data in serenity:next.


FelixMcFelix (approved):
Great, thanks for making these changes! I think this is a strong improvement.

PR #226, opened 2024-02-25
Driver: Support `tokio-websockets`

decahedron1:
Allows for configuring the crate to use [`tokio-websockets`](https://crates.io/crates/tokio-websockets) instead of `tokio-tungstenite` since Twilight 0.16 [is switching to TWS](https://github.com/twilight-rs/twilight/blob/main/book/src/versions/0.16/summary.md#switch-to-tokio-websockets-and-fastrand).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,13 +48,13 @@ jobs:
             os: windows-latest
             dont-test: true
           - name: driver only
-            features: driver rustls
+            features: driver tungstenite rustls
             dont-test: true
           - name: gateway only
-            features: gateway serenity rustls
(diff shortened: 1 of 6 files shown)


Erk- (approved):
Looks good, though maybe it should be retargeted to `next` since I think it is a breaking change to add a new required feature as is done here with `tungstenite`, I have also tested this with a 0.16.0-rc.1 branch and it works there.

I will also defer the decision if `tws` is the feature name we want for this.

FelixMcFelix (approved):
Thanks a lot for putting this together. I can't think of any better flag name than `tws` right now, so all looks good to me!

PR #228, opened 2024-02-27
Remove Simd-json

Erk-:
Depends on #226 

This mirrors the serenity pr https://github.com/serenity-rs/serenity/pull/2735.

### DIFF
--- a/src/input/codecs/dca/mod.rs
+++ b/src/input/codecs/dca/mod.rs
@@ -110,11 +110,10 @@ impl FormatReader for DcaReader {
                 return symph_err::decode_error("missing DCA1 metadata block");
             }
 
-            let mut raw_json = source.read_boxed_slice_exact(size as usize)?;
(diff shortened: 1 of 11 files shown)


src/input/codecs/dca/mod.rs
```diff
@@ -110,11 +110,10 @@ impl FormatReader for DcaReader {
                 return symph_err::decode_error("missing DCA1 metadata block");
             }
 
-            let mut raw_json = source.read_boxed_slice_exact(size as usize)?;
+            let raw_json = source.read_boxed_slice_exact(size as usize)?;
 
-            // NOTE: must be mut for simd-json.
             #[allow(clippy::unnecessary_mut_passed)]
```
FelixMcFelix: I think you can drop this clippy lint now that we never need a `&mut` reference.

FelixMcFelix (approved):
Looks good, thanks very much! I think there's one clippy ignore statement we can be rid of, then it's all good. :)

PR #235, opened 2024-03-27
Make DisposalThread reachable

AlvaroMS25:
The `Config` object provided to `Call`s and `Driver`s allows setting a `DisposalThread`, but since it is unreachable from outside the crate, the only way to properly set it is using a `Songbird` instance, which may not be adequate for all use cases. This PR just makes `DisposalThread` reachable from the outside.

### DIFF
--- a/src/driver/mod.rs
+++ b/src/driver/mod.rs
@@ -64,6 +64,8 @@ use core::{
 use flume::{r#async::RecvFut, SendError, Sender};
 #[cfg(feature = "builtin-queue")]
 use std::time::Duration;
+#[allow(unused_imports)]
+pub use tasks::disposal::DisposalThread;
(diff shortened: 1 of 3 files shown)


FelixMcFelix (commented):
The idea of this makes sense (easier to manage disposal thread to driver ratio in non-gateway use cases). However, I think you've made a lot more public than you intended, which is why the docs lint is flagging up. Would this work instead as a `pub use tasks::disposal::DisposalThread;` in src/driver/mod.rs?

AlvaroMS25:
I'm sorry about that, i didn't realize how much i was making public with that. You're right, just doing a `pub use` would be enough. I'll update this PR in a sec

FelixMcFelix:
Fantastic, thanks very much!

PR #236, opened 2024-03-31
Fix clippy warnings

GnomedDev:
Some changes here seem breaking, but I'm pretty sure they are crate internal and that's why the warnings fired. This most likely needs checking though.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -390,8 +390,8 @@ impl Config {
 
 // Test only attributes
 #[cfg(all(test, feature = "driver"))]
+#[allow(missing_docs)]
 impl Config {
-    #![allow(missing_docs)]
     #[must_use]
     pub fn tick_style(mut self, tick_style: TickStyle) -> Self {
         self.tick_style = tick_style;
--- a/src/driver/bench_internals.rs
+++ b/src/driver/bench_internals.rs
@@ -14,6 +14,7 @@ use crate::{
     tracks::{Track, TrackHandle},
 };
 
+#[must_use]
 pub fn track_context(t: Track) -> (TrackHandle, TrackContext) {
     t.into_context()
 }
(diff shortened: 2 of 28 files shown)


src/error.rs
```diff
@@ -47,10 +47,10 @@ pub enum JoinError {
     /// re-attempting connection.*
     Driver(ConnectionError),
     #[cfg(feature = "serenity")]
-    /// Serenity-specific WebSocket send error.
+    /// Serenity-specific `WebSocket` send error.
```
FelixMcFelix: I've seen this change in a few places: is clippy assuming that CamelCase words must be types/code terms? I'm not too fussed either way on presentation, because I assume the appropriate `#[allow]` would be more obnoxious than just... doing this.

src/input/metadata/mod.rs
```diff
@@ -28,7 +28,7 @@ pub struct AuxMetadata {
 
     /// The number of audio channels in this stream.
     pub channels: Option<u8>,
-    /// The YouTube channel of this stream.
+    /// The `YouTube` channel of this stream.
```
FelixMcFelix: Yeah this one's a bit more clear cut for needing suppressed.

FelixMcFelix (commented):
Looks mostly good, The task error changes are all internal so we should ha ve no issues there. Just some frankly odd code fence suggestions from clippy. :)

src/error.rs
GnomedDev: Yeah, I was planning to go back over this and sort that out because this was done by `clippy --fix`, but if you are okay with it I'll leave it in.

GnomedDev:
Ignored the warning causing `--fix` to add the weird backticks.

FelixMcFelix (approved):
Thanks!

PR #237, opened 2024-03-31
Bump dependencies 

GnomedDev:
This is a breaking change as reqwest is publically exposed via HttpRequest::client and maybe other fields.


FelixMcFelix:
Thanks for keeping an eye on this!

PR #238, opened 2024-04-02
Rewrite basic voice example to use poise

GnomedDev:
WIP.

These rewrites are based on the versions on serenity-next, as they already use the UserData system that was ported from poise to serenity on serenity@next.


nathan2slime:
Thanks for the code, it helped me implement my Cha Hae-In inspired bot

nathan2slime:
It seems that there is a compatibility issue in the `Client` type, I had to downgrade `reqwest `to `version 0.11`

PR #242, opened 2024-07-06
feat(input): Support HLS streams

### LINKED ISSUE #241: M3U8 support (soundcloud)
**Songbird version:** 0.4.1

**Rust version (`rustc -V`):** rustc 1.81.0-nightly (aa1d4f682 2024-07-03)

**Serenity/Twilight version:** serenity 0.12

**Output of `ffmpeg -version`, `yt-dlp --version` (if relevant):** 2024.07.02
...

**Description:**
when trying to play a source from soundcloud.com, symphonia throws a no format supported error, even when enabling everything. my best guess is that soundcloud uses opus or something.
...

**Steps to reproduce:**
...
use voice example
pass in soundcloud link (e.g [https://soundcloud.com/upscale-recordings/tomsbattery?](https://soundcloud.com/upscale-recordings/tomsbattery?))
observe error thrown in logs

Erk-:
I am not quite sure about saying that it is lazy, because it eagerly downloads the stream. And thus it can also be used for livestreams. But it fits better into the Lazy Input type imo.

Closes: #241

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ version = "0.4.1"
 async-trait = { optional = true, version = "0.1" }
 audiopus = { optional = true, version = "0.3.0-rc.0" }
 byteorder = { optional = true, version = "1" }
-bytes = { optional = true, version = "1" }
+bytes = { version = "1" }
 crypto_secretbox = { optional = true, features = ["std"], version = "0.1" }
 dashmap = { optional = true, version = "5" }
 derivative = "2"
@@ -41,6 +41,7 @@ serenity-voice-model = { optional = true, version = "0.2" }
 simd-json = { features = ["serde_impl"], optional = true, version = "0.13" }
 socket2 = { optional = true, version = "0.5" }
 streamcatcher = { optional = true, version = "1" }
+stream_lib = { version = "0.4.1" }
 symphonia = { default_features = false, optional = true, version = "0.5.2" }
 symphonia-core = { optional = true, version = "0.5.2" }
 tokio = { default-features = false, optional = true, version = "1.0" }
@@ -131,7 +132,7 @@ twilight = ["dep:twilight-gateway","dep:twilight-model"]
 
 # Behaviour altering features.
 builtin-queue = []
-receive = ["dep:bytes", "discortp?/demux", "discortp?/rtcp"]
+receive = ["discortp?/demux", "discortp?/rtcp"]
 
 # Used for docgen/testing/benchmarking.
 full-doc = ["default", "twilight", "builtin-queue", "receive"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ serenity-voice-model = { optional = true, version = "0.2" }
 simd-json = { features = ["serde_impl"], optional = true, version = "0.13" }
 socket2 = { optional = true, version = "0.5" }
 streamcatcher = { optional = true, version = "1" }
-stream_lib = { version = "0.4.1" }
+stream_lib = { optional = true, version = "0.4.1" }
 symphonia = { default_features = false, optional = true, version = "0.5.2" }
 symphonia-core = { optional = true, version = "0.5.2" }
 tokio = { default-features = false, optional = true, version = "1.0" }
@@ -99,6 +99,7 @@ driver = [
     "dep:serenity-voice-model",
     "dep:socket2",
     "dep:streamcatcher",
+    "dep:stream_lib",
     "dep:symphonia",
     "dep:symphonia-core",
     "dep:tokio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ version = "0.4.1"
 async-trait = { optional = true, version = "0.1" }
 audiopus = { optional = true, version = "0.3.0-rc.0" }
 byteorder = { optional = true, version = "1" }
-bytes = { version = "1" }
+bytes = { optional = true, version = "1" }
 crypto_secretbox = { optional = true, features = ["std"], version = "0.1" }
 dashmap = { optional = true, version = "5" }
 derivative = "2"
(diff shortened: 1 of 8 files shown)


drac-dot-gay:
works perfectly with soundcloud streams!

Cargo.toml
```diff
@@ -17,7 +17,7 @@ version = "0.4.1"
 async-trait = { optional = true, version = "0.1" }
 audiopus = { optional = true, version = "0.3.0-rc.0" }
 byteorder = { optional = true, version = "1" }
-bytes = { optional = true, version = "1" }
+bytes = { version = "1" }
```
FelixMcFelix: This should remain optional, see below.

Cargo.toml
```diff
@@ -41,6 +41,7 @@ serenity-voice-model = { optional = true, version = "0.2" }
 simd-json = { features = ["serde_impl"], optional = true, version = "0.13" }
 socket2 = { optional = true, version = "0.5" }
 streamcatcher = { optional = true, version = "1" }
+stream_lib = { version = "0.4.1" }
```
FelixMcFelix: This should be optional, on driver. See below.

Cargo.toml
```diff
@@ -131,7 +132,7 @@ twilight = ["dep:twilight-gateway","dep:twilight-model"]
 
 # Behaviour altering features.
 builtin-queue = []
-receive = ["dep:bytes", "discortp?/demux", "discortp?/rtcp"]
+receive = ["discortp?/demux", "discortp?/rtcp"]
```
FelixMcFelix: I think we want to keep this as is, and add

```rust
driver = [
    #...
    "dep:bytes",
    #...
    "dep:stream_lib",
    #...
]
```

(Can't put a review comment on unchanged code...)

src/input/sources/hls.rs
```diff
@@ -0,0 +1,139 @@
...
+    io::{ErrorKind as IoErrorKind, Result as IoResult, SeekFrom},
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use bytes::Bytes;
+use futures::StreamExt;
+use pin_project::pin_project;
+use reqwest::{header::HeaderMap, Client};
+use serenity::async_trait;
```
FelixMcFelix:
```suggestion
use async_trait::async_trait;
```

Songbird can be built as a pure audio server, so serenity isn't always a valid dependency.

src/input/sources/hls.rs
```diff
@@ -0,0 +1,139 @@
...
+impl AsyncMediaSource for HlsStream {
+    fn is_seekable(&self) -> bool {
+        false
+    }
+
+    async fn byte_len(&self) -> Option<u64> {
+        None
+    }
+
+    async fn try_resume(
```
FelixMcFelix, line 113: If we're not getting any issues with playback at the moment, then this is probably fine to leave unimplemented for now. Seemingly YouTube are the only folks who like to terminate a stream at 95%.

src/input/sources/hls.rs
```diff
@@ -0,0 +1,139 @@
...
+    Compose,
+    Input,
+};
+
+/// Lazy HLS stream
+#[derive(Debug)]
+pub struct HlsRequest {
+    /// HLS downloader
+    pub hls: Option<DownloadStream>,
+}
```
FelixMcFelix: I think this would be closer to the intent of `Compose` if `HlsRequest` just held onto a `Client` and `String` like HttpRequests do -- this is generally fine since reqwest clients are internally `Arc`'d. `Compose` structs are meant as a 'blueprint' of sorts to create usable streams. That way you create a `DownloadStream`/`HlsStream` every time you call `HlsRequest::create_async`.

This also prevents the 'can only use once' issue below.

src/input/sources/hls.rs
```diff
@@ -0,0 +1,139 @@
...
+        HlsRequest { hls: Some(hls) }
+    }
+
+    fn create_stream(&mut self) -> Result<HlsStream, AudioStreamError> {
+        let Some(hls) = self.hls.take() else {
+            return Err(AudioStreamError::Fail("hls can only be used once".into()));
+        };
+
+        let stream = Box::new(StreamReader::new(hls.map(|ev| match ev {
+            Event::Bytes { bytes } => Ok(bytes),
```
FelixMcFelix, line 65: I guess a question on how `stream_lib` works: is it downloading all chunks in sequence in the background and holding onto them, or are you doing something like a fixed-size lookahead? I'm really just curious here, since the cost of holding an Opus-encoded song in memory is like <10MB which I'd consider acceptable for enabling HLS support.

src/input/sources/hls.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    async fn create_async(
+        &mut self,
+    ) -> Result<AudioStream<Box<dyn MediaSource>>, AudioStreamError> {
+        Err(AudioStreamError::Unsupported)
+    }
+
+    fn should_create_async(&self) -> bool {
+        false
```
FelixMcFelix, line 141: Likely `true` if we go with the suggested changes.

FelixMcFelix (commented):
Thanks a lot for taking this work on, and for your existing development into `stream_lib`! I just have some questions around dependencies and fitting better into the `Compose`/`Lazy` framework.

src/input/sources/hls.rs
Erk-: So at the moment it just downloads all chunks in the hls playlist and queue them in, as they are put into bytes they are reference counted so I think they should be dropped as they are played.

I will have to think about it a bit to figure out how to do a fixed size lookahead (either time or size based) but it is something that could be added.

The place where it happens: https://github.com/Erk-/rsget/blob/master/stream_lib/src/hls/mod.rs#L117

(The code could use a clean up, looking at that file now I can see it is quite messy)

src/input/sources/hls.rs
Erk-: DownloadStream does not need async to be created so this is still false.

FelixMcFelix (approved):
Great, I think this looks to go! Feel free to Squash & Merge when you're ready.

PR #243, opened 2024-07-07
fix(conn): Some operating systems does not allow 0 recv buffer

Erk-:
This will cause FreeBSD to fail setting up the socket. It may also be true of some other operating systems, but these are the ones I have been able to test.

It works on Windows and Linux.


FelixMcFelix (approved):
Good catch, and simple enough fix. Thank you!

PR #244, opened 2024-07-09
fix(dependecies): don't use default features for stream_lib

Erk-:
stream_lib would always pull in native-tls, this was changed in 0.4.2
which contains a backport of https://github.com/Erk-/rsget/pull/45.


FelixMcFelix (approved):
Good catch, thanks!

PR #249, opened 2024-08-21
Fix(receive): Handle sleds of identical RTP timestamps

FelixMcFelix:
Bots joining calls with users seem to provoke large runs of packets with identical timestamps -- the existing logic was intended to handle this catchup case in addition to the normal (+=960) at all times.

However, we were checking that a packet was modulo greater-than the next ts, rather than modulo less-than. Simple enough to fix.


FelixMcFelix:
The user who encountered this bug reported the fix as working in Discord. CI failures are unrelated; I'll hack together something to skip those tests on CI shortly.

PR #256, opened 2024-09-28
Expose an AsyncReadOnlySource

rhgndf:
`HlsStream` already wraps any struct with the `AsyncRead` trait into an `AsyncMediaSource`, so rename it to `AsyncReadOnlySource` and expose it in the crate. This is very similar to the sync version `symphonia_core::io::ReadOnlySource`.

A possible use case would be streaming PCM samples to a discord voice channel:
```
let (audio_receiver, mut audio_sender) = tokio::io::simplex(76800);
let audio_stream = AsyncAdapterStream::new(
    Box::new(AsyncReadOnlySource::new(audio_receiver)),
    76800,
);
let adapter = RawAdapter::new(audio_stream, 48000, 2);
let _ = handler.play_input(adapter.into());
```

Unfortunately, doing `cargo make ready` results in a lot of errors and I am not familiar enough with the codebase to add docs or fix them. However, since this is mostly shuffling around existing code, there shouldn't be any additional errors.

### DIFF
--- a/src/input/sources/hls.rs
+++ b/src/input/sources/hls.rs
@@ -1,22 +1,14 @@
-use std::{
-    io::{ErrorKind as IoErrorKind, Result as IoResult, SeekFrom},
-    pin::Pin,
-    task::{Context, Poll},
-};
-
 use async_trait::async_trait;
 use bytes::Bytes;
 use futures::StreamExt;
-use pin_project::pin_project;
 use reqwest::{header::HeaderMap, Client};
 use stream_lib::Event;
 use symphonia_core::io::MediaSource;
-use tokio::io::{AsyncRead, AsyncSeek, ReadBuf};
 use tokio_util::io::StreamReader;
 
 use crate::input::{
     AsyncAdapterStream,
-    AsyncMediaSource,
+    AsyncReadOnlySource,
     AudioStream,
     AudioStreamError,
     Compose,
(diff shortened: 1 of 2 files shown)


Makefile.toml
```diff
@@ -30,7 +30,7 @@ command = "cargo"
 dependencies = ["format"]
 
 [tasks.build-variants]
-dependencies = ["build", "build-gateway", "build-driver", "build-simd"]
+dependencies = ["build", "build-gateway", "build-driver"]
```
FelixMcFelix: Can I ask why you've done this? `async-simd` is still supported as of 0.4.x.

FelixMcFelix (commented):
At first glance this looks reasonable enough to me; thanks for putting it together. `AsyncReadOnlySource` is a useful utility to expose.

One minor nit/question. Otherwise, I would ignore the CI failure on missing documentation. Those are benchmark-harness specific methods which should be marked s.t. the `missing_docs` lint skips them. I'm not sure what's changed in nightly to cause this, but I'll tackle that shortly.

Makefile.toml
rhgndf: This makes some tests run when rebased on top of serenity-next. Probably due to this commit:

https://github.com/serenity-rs/songbird/commit/05cf36b040ff7c4c62be707dd83a2ec7962168b3#diff-9375fd04332c86472d7be397ef09428cb86babd8826880a5835bd1d1c1bdbc08L11

I have removed the commit.

FelixMcFelix: Yeah, the support is removed from `next` but still present here. Thanks for fixing that up.

FelixMcFelix (approved):
Looks good to me, thanks; happy that this is non-breaking since most of the `HlsStream` internals are private. Feel free to merge when you're ready!

EDIT: D'oh, I'll pull it in.

PR #262, opened 2024-10-29
fix(source): HLS would not work if used directly.

### LINKED ISSUE #257: HLS stream would panic
**Songbird version:** 0.4.3

**Rust version (`rustc -V`):** rustc 1.81.0 (eeb90cda1 2024-09-04)

**Description:**

The program panicked in rusty pool thread when using `HlsRequest`

```
thread 'rusty_pool_1_thread_1' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/stream_lib-0.5.2/src/hls/mod.rs:99:9:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

**Steps to reproduce:**

```rust
let input = HlsRequest::new(...).into();
driver.play_only_input(input);
```

Erk-:
This changes HlsRequest to return true in
Compose::should_create_async.

This would otherwise cause it to be spawned on a thread without a
executor causing it to panic.

This would only cause issues if used directly since we otherwise use
Compose::create in a context where we have access to the executor.

Closes #257 

## Open questions:

- [x] Should the `create` implementation be removed or have a note that it will panic if called outside of a tokio executor?

### DIFF
--- a/src/input/sources/hls.rs
+++ b/src/input/sources/hls.rs
@@ -16,6 +16,11 @@ use crate::input::{
 };
 
 /// Lazy HLS stream
+///
+/// # Note:
+///
+/// `Compose::create` for this struct panics if called outside of a
+/// tokio executor since it uses background tasks.
 #[derive(Debug)]
 pub struct HlsRequest {
     /// HTTP client
@@ -82,11 +87,11 @@ impl Compose for HlsRequest {
     async fn create_async(
         &mut self,
     ) -> Result<AudioStream<Box<dyn MediaSource>>, AudioStreamError> {
-        Err(AudioStreamError::Unsupported)
+        self.create()
     }
 
     fn should_create_async(&self) -> bool {
-        false
+        true
     }
 }
 


Erk-:
The fails in the tests seems to be that beta and nightly catches more examples of missing documentation and is not caused by this patch.

kane50613 (approved):

FelixMcFelix (approved):
Looks good to me, thanks for figuring this out. Yeah, CI failures are down to a change in how a lint is applied.

FelixMcFelix:
>     * [ ]  Should the `create` implementation be removed or have a note that it will panic if called outside of a tokio executor?

I think a note indicating the panic would be most useful here. It is still feasible to call, so it doesn't violate the existing API contract but it's better to not surprise users where possible!

Erk-:
I added a note about it on the struct

FelixMcFelix:
Thanks for the fix; feel free to merge when you're ready (I believe you have access?).

PR #265, opened 2024-11-14
Receive: Config of decode sample rate/channels

FelixMcFelix:
This PR allows for dynamic configuration of the output sample rate and channel count of received Opus audio. Users who rely on supported formats should no longer need to manually resample & downmix audio decoded from SSRCs in a call.

Opus exposes tuples of (Mono, Stereo) x (8, 12, 16, 24, 48)kHz. Changing this at runtime (mid-call) may cause some audio glitches, as decoder state must be reconstructed from scratch for all affected SSRCs.

### DIFF
--- a/src/driver/decode_mode.rs
+++ b/src/driver/decode_mode.rs
@@ -35,7 +35,7 @@ pub enum Channels {
     Mono,
     /// Decode received audio packets into two interleaved channels.
     ///
-    /// Received mono packets will automatically be duplicated across
+    /// Received mono packets' samples will automatically be duplicated across
     /// both channels.
     ///
     /// The default choice.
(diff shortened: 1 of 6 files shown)


src/driver/decode_mode.rs
```diff
@@ -24,3 +26,68 @@ impl DecodeMode {
...
+/// The sample rate of output audio when using [`DecodeMode::Decode`].
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash)]
+#[non_exhaustive]
+pub enum SampleRate {
+    /// Decode to a sample rate of 8kHz.
+    Hz8000,
+    /// Decode to a sample rate of 12kHz.
+    Hz12000,
+    /// Decode to a sample rate of 46kHz.
+    Hz16000,
```
tazz4843, line 73: Just noticed 46kHz in the comment vs value of 16000Hz here

FelixMcFelix: Good catch!

Erk- (approved):

PR #266, opened 2024-11-18
Receive: Fix handling for large timestamp/sequence jumps

### LINKED ISSUE #261: Possible current_timestamp desynchronization in `PlayoutBuffer`
**Songbird version:** git serenity-next

**Rust version (`rustc -V`):** `rustc 1.82.0-nightly (176e54520 2024-08-04)`

**Serenity/Twilight version:** git next

**Description:**
Scripty has been having issues with VoiceTick returning empty audio ever since a few weeks before DAVE introduction (possibly related, not worth digging into yet I don't think). Digging through logs, I noticed lots of lines noting `Witholding packet: ts_diff is -xxxxxx` or `Packet arrived beyond playout max length: wanted slot nn.` I was going to open a bug report long ago but I finally got around to it.

[Explore-logs-2024-10-29 05_34_55.txt](https://github.com/user-attachments/files/17552164/Explore-logs-2024-10-29.05_34_55.txt)

Attached is a log file from the most recent known event of a desync. The first line where audio seems to disappear is `Witholding packet: ts_diff is -403200`. The timestamp difference between that and the last audio packet from this speaker approximately nine seconds before is 412800, 9600 (or `10*MONO_FRAME_SIZE`) off of the actual ts_diff value. Based off this very basic poking and prodding I suspect a possible race condition somewhere, but don't know enough about songbird internals to poke into it much.

There's been lots of comments from me in the serenity discord about this so I'll copy paste some of them here for a full thread: whatever follows is copied+pasted

> there's at least three possible failure states i know of right now
> * VoiceTick is always silent for every user no matter what
> * VoiceTick returns audio for a handful of users for an arbitrary length of time and then goes silent forever for all
> * VoiceTick always returns data for some users but never others
> 
> and as far as i know the failure state varies from session to session
> during all of this the WebSocket is unaffected and i still get speaker updates normally no matter what which triggers empty results getting sent to scripty's STT service which return unusually fast (that's the thing that originally tipped me off here) 

I'm willing to put a $50 USD bounty on this issue, through GitHub Sponsors or kofi. Let me know fi any additional info or anything else is required to help debug :)

FelixMcFelix:
This PR addresses some issues which have cropped up on voice receive at scale:

* In unknown circumstances, we can be left with adjacent packets queued which have very different timestamps. The playout buffer would withhold its held packets, leading to the loss of many subsequent packets if the timestamp jump is larger than 64 frames. This seems to occur for some specific clients which join before a bot, suggesting the DAVE -> legacy switchover is involved.
* Some loss patterns can leave us unable to correctly track the next expected sequence number (i.e., large loss runs), leaving the playout buffer unable to accept any packets if the packet sequence differed by over 64 entries.

The fixes are fallbacks which treat sufficiently large desynchronisation, and allow the playout to get back into a consistent state in both cases. Large timestamp jumps on adjacent packets now update the next expected TS (noting that we only want to withhold a few playout delays at most ). Failure to insert 0.25s of packets (or attempting to add a new sequence number into an empty buffer) can now take precedence.

Closes #261.

### DIFF
--- a/src/driver/tasks/udp_rx/playout_buffer.rs
+++ b/src/driver/tasks/udp_rx/playout_buffer.rs
@@ -45,6 +45,7 @@ pub struct PlayoutBuffer {
     playout_mode: PlayoutMode,
     next_seq: RtpSequence,
     current_timestamp: Option<RtpTimestamp>,
+    consecutive_store_fails: usize,
 }
 
 impl PlayoutBuffer {
@@ -54,6 +55,7 @@ impl PlayoutBuffer {
             playout_mode: PlayoutMode::Fill,
             next_seq,
             current_timestamp: None,
+            consecutive_store_fails: 0,
         }
     }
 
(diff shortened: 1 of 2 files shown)


FelixMcFelix:
The main bot encountering these issues has had no recurrence after applying these fixes, so they appear good to go.

PR #268, opened 2024-11-22
Accept optional additional user arguments for `yt-dlp`

Henauxg:
This small PR makes it possible to give additional user arguments to the `yt-dlp` process from the code.

Feel free to close the PR if this solution does not suit you.


FelixMcFelix (approved):
Thanks very much -- this looks straightforward and good to me.

Issue #232, opened 2024-03-04
Loops stop working after stopping or playing audio over another


strinnityk:
**Songbird version:** 0.4.1

**Rust version (`rustc -V`):** rustc 1.75.0 (82e1608df 2023-12-21)

**Serenity version:** 0.12.1

**Output of `ffmpeg -version`, `yt-dlp --version` (if relevant):**
Not using either of those, but just in case:
-  ffmpeg version 6.0

**Description:**
Loops stop working after running `stop` or playing another audio when one is already being played. 
This happens both for `LoopState.Infinite`, as well as `LoopState.Finite(x)` for `x > 0`.

**Steps to reproduce:**
1. Instantiate two tracks with `LoopState.Infinite`. They can be the same file.
2. Play one of them.
3. While the previous track is playing, play the other one.
4. The currently playing track while get overridden by the new one, but the new one will behave as `LoopState.Finite(0)`, not as expected with `LoopState.Infinite`.
5. From now on, `LoopState.Infinite` doesn't work anymore. All tracks played after this will behave as if `LoopState.Finite(0)`.

**Additional Information** 
There are many other combinations that come to my mind, but I haven't tried them all. Let me know if I can help debugging this in any way.

FelixMcFelix:
It sounds as though there may be an ordering mismatch between tracks in the event context and the audio task itself. I don't know when I'll have the time to dig into this, but if you add an event listener (state change, loop, etc.) you should be able to see what the Event side of the house believes your track state is. You can compare this with the output from [`TrackHandle::get_info`](https://docs.rs/songbird/latest/songbird/tracks/struct.TrackHandle.html#method.get_info) to see who's wrong.

strinnityk:
> It sounds as though there may be an ordering mismatch between tracks in the event context and the audio task itself. I don't know when I'll have the time to dig into this, but if you add an event listener (state change, loop, etc.) you should be able to see what the Event side of the house believes your track state is. You can compare this with the output from [`TrackHandle::get_info`](https://docs.rs/songbird/latest/songbird/tracks/struct.TrackHandle.html#method.get_info) to see who's wrong.

Thanks for the answer. I'll give it a try in the weekend :)

ofiddy:
I've had this problem too. I tried having a look at it myself but I'm definitely too green with Rust to figure out what was going on. Only the first song played would loop: any others would not.

JellyWX:
[...]

PR #270, opened 2024-12-03
Clear track_handles when SetTrack heard

JellyWX:
## Summary

Clear `track_handles` when `SetTrack` heard, to prevent issue where indexes of `tracks` become desynced with indexes of `track_handles`. See #232 

## Testing

Not done, waiting for confirmation that this is a sensible fix


FelixMcFelix (approved):
That looks believable, and it looks like exactly the sort of mistake that underlies that bug. 🤦  Great catch!

PR #272, opened 2025-02-22
Move decode config into DecodeMode::Decode

GnomedDev:
This better models the configuration and shrinks `UdpRxMessage` significantly.


FelixMcFelix (approved):
Nice, thanks for reorganising there. Glad to cut down the message size there.

PR #274, opened 2025-03-21
Fix breaking change from removing ShardRunnerMessage::Message

NateDiamond13:
Fixed the compiler issue introduced in the "next" branch of serenity with the removal of the Message variant of ShardRunnerMessage enum in [e552918](https://github.com/serenity-rs/serenity/commit/e552918da9f976af2684b544eddbb13796d34599).

### DIFF
--- a/src/shards.rs
+++ b/src/shards.rs
@@ -173,17 +173,14 @@ impl VoiceUpdate for Shard {
         match self {
             #[cfg(feature = "serenity")]
             Shard::Serenity(handle) => {
-                let map = json!({
-                    "op": 4,
-                    "d": {
-                        "channel_id": channel_id.map(ChannelId::get),
-                        "guild_id": guild_id.get(),
-                        "self_deaf": self_deaf,
-                        "self_mute": self_mute,
-                    }
-                });
-
-                handle.send(ShardRunnerMessage::Message(map.to_string().into()))?;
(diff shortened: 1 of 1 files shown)


GnomedDev (approved):
Looks good to merge, can you fix the doc CI failure or should I later?

NateDiamond13:
Cool, I'll let you handle the CI doc stuff when you get the chance

src/shards.rs
```diff
@@ -173,17 +173,14 @@ impl VoiceUpdate for Shard {
...
-                        "self_mute": self_mute,
-                    }
-                });
-
-                handle.send(ShardRunnerMessage::Message(map.to_string().into()))?;
+                handle.send(ShardRunnerMessage::UpdateVoiceState {
+                    guild_id: guild_id.get().into(),
+                    channel_id: channel_id
+                        .map(ChannelId::get)
+                        .map(serenity::all::ChannelId::from),
```
mkrasnitski, line 180: Why the roundtrip from `ChannelId` to `u64` and back again?

FelixMcFelix: Songbird defines its own ID newtypes to be independent of whichever gateway a user wants, and the input is in our ID format. We have a From impl from serenity and twilight types to ours, but not apparently the opposite direction (hence the transformation here, I guess).

PR #275, opened 2025-03-27
Fix clippy warnings

GnomedDev:

PR #276, opened 2025-03-27
Properly fix clippy type size warnings

GnomedDev:
This fixes the warnings that were allowed in #275.

This also removes the unused hint field in AudioStream.


src/events/store.rs
```diff
@@ -15,7 +16,7 @@ use tracing::info;
 /// [`EventData`]: EventData
 pub struct EventStore {
     timed: BinaryHeap<EventData>,
-    untimed: HashMap<UntimedEvent, Vec<EventData>>,
+    untimed: Box<EnumMap<UntimedEvent, Vec<EventData>>>,
```
Erk-, line 19: Does it make sense to put this on the heap? You save 64 bit compared to just keeping it on the stack. (at least if I understand the EnumMap implementation this will be a array of size 2)

Although it is still more than one pointer smaller than a HashMap.

Erk- (approved):
Looks good, I have one small question, and I think that the changes related to hint should probably be noted in the pr/merged commit message.

src/events/store.rs
GnomedDev: Nope, this isn't an array of two, since it recurses into the TrackEvent and CoreEvent enums which are much larger, from what I can remember at least.

PR #279, opened 2025-05-04
Playlist fixes, add `YoutubeDl::get_stream` function

### LINKED ISSUE #277: Feature request: Make songbird::input::metadata::ytdl::Output public
Hi there,

I find myself in a situation where I would like to parse the output of my own call to yt-dlp to a AuxMetadata and I realized that `songbird::input::metadata::ytdl::Output` already implements all necessary deserialization and conversion for doing that but unfortunately it it not available outside of the crate. Would you have any strong objections against making this struct public?

Thanks!

ImSapphire:
Example usage of `YoutubeDl::get_stream`: https://codeberg.org/sapphiree/poise-music-bot/src/commit/3b06c53a32ac2657c36648205c00b75fc55a62ce/src/main.rs#L205
My main motivation for making these changes is that adding a full playlist was *really* slow before due to not having access to the stream URL, so I had to re-invoke yt-dlp for every entry. I don't believe this will break existing code.
The doc comments are mostly copied from other places and are probably insufficient, sorry about that.

Also closes #277.

### DIFF
--- a/src/input/metadata/mod.rs
+++ b/src/input/metadata/mod.rs
@@ -1,9 +1,11 @@
+//! Metadata about outputs - currently only [`ytdl::Output`] is publicly exported.
+
 use crate::error::JsonError;
 use std::time::Duration;
 use symphonia_core::{meta::Metadata as ContainerMetadata, probe::ProbedMetadata};
 
 pub(crate) mod ffprobe;
-pub(crate) mod ytdl;
+pub mod ytdl;
 
 use super::Parsed;
 
--- a/src/input/metadata/mod.rs
+++ b/src/input/metadata/mod.rs
@@ -1,11 +1,12 @@
-//! Metadata about outputs - currently only [`ytdl::Output`] is publicly exported.
+//! Metadata formats specific to [`Compose`] types.
 
 use crate::error::JsonError;
 use std::time::Duration;
 use symphonia_core::{meta::Metadata as ContainerMetadata, probe::ProbedMetadata};
 
 pub(crate) mod ffprobe;
-pub mod ytdl;
+mod ytdl;
+pub use ytdl::Output as YoutubeDlOutput;
 
 use super::Parsed;
 
--- a/src/input/metadata/mod.rs
+++ b/src/input/metadata/mod.rs
@@ -1,4 +1,4 @@
-//! Metadata formats specific to [`Compose`] types.
+//! Metadata formats specific to [`crate::input::Compose`] types.
 
 use crate::error::JsonError;
 use std::time::Duration;
--- a/src/input/metadata/ytdl.rs
+++ b/src/input/metadata/ytdl.rs
@@ -1,4 +1,4 @@
-//! YoutubeDl track metadata - see [`Output`].
+//! `YoutubeDl` track metadata.
 
 use super::AuxMetadata;
 use crate::constants::SAMPLE_RATE_RAW;
@@ -16,7 +16,7 @@ pub struct Output {
     pub album: Option<String>,
     /// The channel name.
     pub channel: Option<String>,
-    /// The duration of the stream.
+    /// The duration of the stream in seconds.
     pub duration: Option<f64>,
     /// The size of the stream.
     pub filesize: Option<u64>,
@@ -24,9 +24,9 @@ pub struct Output {
     pub http_headers: Option<HashMap<String, String>>,
     /// Release date of this track.
     pub release_date: Option<String>,
-    /// The track thumbnail.
+    /// The thumbnail URL for this track.
     pub thumbnail: Option<String>,
-    /// The title of the track.
+    /// The title of this track.
     pub title: Option<String>,
     /// The track name.
     pub track: Option<String>,
--- a/src/input/metadata/ytdl.rs
+++ b/src/input/metadata/ytdl.rs
@@ -1,28 +1,49 @@
+//! YoutubeDl track metadata - see [`Output`].
+
 use super::AuxMetadata;
 use crate::constants::SAMPLE_RATE_RAW;
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, time::Duration};
 
+/// Information returned by yt-dlp about a URL. 
+/// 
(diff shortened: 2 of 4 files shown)


ImSapphire:
Just now realising I don't even use `get_streams` and it seems a bit redundant, not sure if removing it would be preferred

ImSapphire:
@FelixMcFelix Is there a chance you could have a peek at this?

FelixMcFelix:
> @FelixMcFelix Is there a chance you could have a peek at this?

I will, hopefully in the first half of next week; I've been busy this week with work, and this weekend's fairly limited.

src/input/metadata/mod.rs
```diff
@@ -1,9 +1,11 @@
+//! Metadata about outputs - currently only [`ytdl::Output`] is publicly exported.
```
FelixMcFelix:
```suggestion
//! Metadata formats specific to [`Compose`] types.
```

src/input/mod.rs
```diff
@@ -59,7 +59,7 @@ mod error;
 #[cfg(test)]
 pub mod input_tests;
 mod live_input;
-mod metadata;
+pub mod metadata;
```
FelixMcFelix, line 62: I can't comment on code you haven't touched, but below we have:

```rust
pub use self::{
    adapters::*,
    audiostream::*,
    compose::*,
    error::*,
    live_input::*,
    metadata::*,
    parsed::*,
    sources::*,
};
```

I think we want to go for:

```rust
pub use self::{
    adapters::*,
    audiostream::*,
    compose::*,
    error::*,
    live_input::*,
    metadata::{AuxMetadata, Metadata},
    parsed::*,
    sources::*,
};
```

I know those are reachable via `metadata`, but we do want them to be signalled as the 'main' types (and not pollute the `input` namespace with `ytdl` and friends).

src/input/metadata/ytdl.rs
```diff
@@ -1,28 +1,49 @@
+//! YoutubeDl track metadata - see [`Output`].
```
FelixMcFelix:
```suggestion
//! `YoutubeDl` track metadata.
```

src/input/metadata/mod.rs
```diff
@@ -1,9 +1,11 @@
...
+
 use crate::error::JsonError;
 use std::time::Duration;
 use symphonia_core::{meta::Metadata as ContainerMetadata, probe::ProbedMetadata};
 
 pub(crate) mod ffprobe;
-pub(crate) mod ytdl;
+pub mod ytdl;
```
FelixMcFelix: I'd be tempted to recast this as:

```suggestion
mod ytdl;
pub use ytdl::Output as YoutubeDlMetadata;
```

Will need some fixup in `songbird\src\input\sources\ytdl.rs`, but I think that makes it easier to reach as a user.

src/input/metadata/ytdl.rs
```diff
@@ -1,28 +1,49 @@
...
+/// Returned by [`crate::input::YoutubeDl::query`].
 #[derive(Deserialize, Serialize, Debug)]
 pub struct Output {
+    /// The main artist.
     pub artist: Option<String>,
+    /// The album name.
     pub album: Option<String>,
+    /// The channel name.
     pub channel: Option<String>,
+    /// The duration of the stream.
```
FelixMcFelix:
```suggestion
    /// The duration of the stream in seconds.
```

src/input/metadata/ytdl.rs
```diff
@@ -1,28 +1,49 @@
...
     pub duration: Option<f64>,
+    /// The size of the stream.
     pub filesize: Option<u64>,
+    /// Required HTTP headers to fetch the track stream.
     pub http_headers: Option<HashMap<String, String>>,
+    /// Release date of this track.
     pub release_date: Option<String>,
+    /// The track thumbnail.
```
FelixMcFelix:
```suggestion
    /// The thumbnail URL for this track.
```

src/input/sources/ytdl.rs
```diff
@@ -118,19 +117,16 @@ impl<'a> YoutubeDl<'a> {
...
+        Ok(self
+            .query(n_results)
+            .await?
+            .into_iter()
+            .map(|v| v.as_aux_metadata()))
     }
 
-    async fn query(&mut self, n_results: usize) -> Result<Vec<Output>, AudioStreamError> {
+    /// Runs a search for the given query, returning a list of up to `n_results`
+    /// possible matches which are [`Output`] objects containing a valid URL.
```
FelixMcFelix:
```suggestion
    /// Runs a search for the given query, returning a list of up to `n_results`
    /// possible matches.
```

A valid URL is an implicit part of the output type here. No need to re-specify `[Output]`, either.

FelixMcFelix (commented):
Thanks, I think this generally looks good. Just some tidying around organisation/docs.

src/input/mod.rs
FelixMcFelix: Sorry, I think I was unclear in this comment. I think that we should keep `pub mod metadata;`, and use that to reach `YoutubeDlOutput` (i.e., we have `songbird::input::AuxMetadata`, `songbird::input::Metadata`, and `songbird::input::metadata::YoutubeDlOutput`).

FelixMcFelix:
@ImSapphire Thanks, just one nit (and you need to rerun formatting). Otherwise, the test failure looks like a flake.

ImSapphire:
@FelixMcFelix Formatting should be good now

src/input/mod.rs
ImSapphire: > Sorry, I think I was unclear in this comment. I think that we should keep `pub mod metadata;`, and use that to reach `YoutubeDlOutput` (i.e., we have `songbird::input::AuxMetadata`, `songbird::input::Metadata`, and `songbird::input::metadata::YoutubeDlOutput`).

Oops, I'll do that now

FelixMcFelix (approved):
Thanks for pulling that through -- looks good to me!

PR #281, opened 2025-05-04
Driver: Update the head and tail offsets returned from decrypt_*_in_place().

### LINKED ISSUE #280: Incorrect offsets in RtpPacket
**Songbird version:** 0.5

**Rust version (`rustc -V`):** rustc 1.86.0-nightly (48a426eca 2025-01-12)

**Serenity/Twilight version:** Serenity 0.12.4
...

**Description:** `payload_offset` constantly returns a start offset larger than the payload itself, appears to be `payload.len() + 12`. `payload_end_pad` seems to be returning 16, when it should be returning 20 (to account for the nonce).
...

**Steps to reproduce:**
In an event handler for `Ctx::RptPacket(rpt_data)`:
```rust
let start = rtp_data.payload_offset;
let tail = rtp_data.payload_end_pad;
let rtp = rtp_data.rtp().from_packet();
debug!("Payload has size of {}, with start of {start} and tail of {tail}", rtp.payload.len());
```
...

astroftl:
`payload_offset` is derived from `start_estimate`, which was calculated first taking the start estimate from `decrypt_pkt_in_place()`.

This value was calculated by jumping to the end of the extension header (based on it's fixed minimum size), and then splitting at another offset, `mode.payload_prefix_len()`. However, this only accounts for encryption prefixes, which are zero. Meaning, when `decrypt_pkt_in_place()` returns `plaintext_end + pre_payload.len()` (where `pre_payload.len()` is 0), the returned start offset begins at the end of the extension header, but at the beginning of the extension itself, not the Opus data.

Now, this may be intended; the docs say that the value returns the point where headers end. However, without also including the length of the extension data itself, this offset into the buffer isn't very useful. It may be worth having additional design discussion about the implications of this, and the intended usage of `payload_offset`.

Regardless, because the `start_estimate` returned from `decrypt_pkt_in_place()` is after the headers, later in `decrypt_rtp_in_place()` where `start_estimate` is used to split the payload, the result buffer `exts_and_opus` does not include the extension headers. This means that pnet does not have the header information to reconstruct the actual `payload` length (vs. `ext_data`, since the length field telling where to split the two is missing). This means the value which is added to `start_estimate`, `extension.packet().len() - extension.payload().len()`, is the length of the entire `exts_and_opus` buffer, putting the actual returned `payload_offset` far too large.

My solution is to not perform a split based on the return value of `decrypt_pkt_in_place()` at all. (It felt wrong to change around the body of `decrypt_pkt_in_place()`, since the split point happening after the extension header but before extension data is technically correct for decrypting ciphertext.) Instead, after decryption, I build a `RtpExtensionPacket` from the entire `packet.payload()`, which starts at the beginning of the extension headers. This allows for correct calculation of `payload_offset` from `extension.packet().len() - extension.payload().len()`. This value is then used directly instead of being added to a base value, or returned as 0 in the case of no extensions.

---

`payload_end_pad` is returned from `decrypt_rtp_in_place()`, which is just passing through the value from `decrypt_pkt_in_place()`. This was `post_payload.len()`.

The problem is that `post_payload` is gotten from splitting `body_remaining` at `suffix_split_point`, which is `body_remaining.len() - mode.tag_suffix_len()`; but `body_remaining` is the leftovers after `mode.nonce_slice()` is performed.

Thus, the size of `body_remaining` is `sizeof(nonce)` short. This error propagates through to `post_payload`, making the `post_payload.len()` return value incorrect; it should have the size of the nonce value added back in.

My solution for this was simply to return `post_payload.len() + slice_to_use.len()`.

---

Additionally, since `split_at_checked()` was only used to derive `exts_and_opus`, which I no longer use, and because it was already marked temporary, I've removed it.

---

`cargo make ready` passes all tests.

Resolves #280.

### DIFF
--- a/src/driver/crypto.rs
+++ b/src/driver/crypto.rs
@@ -412,8 +412,8 @@ impl Cipher {
         // Update the start estimate to account for bytes occupied by extension headers.
         let payload_offset = if has_extension {
             let payload = packet.payload();
-            let extension = RtpExtensionPacket::new(payload)
-                .ok_or(InternalError::IllegalVoicePacket)?;
+            let extension =
+                RtpExtensionPacket::new(payload).ok_or(InternalError::IllegalVoicePacket)?;
             extension.packet().len() - extension.payload().len()
         } else {
             0
@@ -484,7 +484,10 @@ impl Cipher {
             },
         }
 
-        Ok((plaintext_end + pre_payload.len(), post_payload.len() + slice_to_use.len()))
+        Ok((
+            plaintext_end + pre_payload.len(),
+            post_payload.len() + slice_to_use.len(),
+        ))
     }
 }
 
--- a/src/driver/crypto.rs
+++ b/src/driver/crypto.rs
@@ -407,19 +407,19 @@ impl Cipher {
             0
         };
 
-        let (mut start_estimate, end) = self.decrypt_pkt_in_place(packet, plain_bytes)?;
+        let (_, end) = self.decrypt_pkt_in_place(packet, plain_bytes)?;
 
         // Update the start estimate to account for bytes occupied by extension headers.
-        if has_extension {
-            let packet = packet.packet();
-            if let Some((_, exts_and_opus)) = split_at_checked(packet, start_estimate) {
-                let extension = RtpExtensionPacket::new(exts_and_opus)
-                    .ok_or(InternalError::IllegalVoicePacket)?;
-                start_estimate += extension.packet().len() - extension.payload().len();
-            }
-        }
+        let payload_offset = if has_extension {
+            let payload = packet.payload();
+            let extension = RtpExtensionPacket::new(payload)
+                .ok_or(InternalError::IllegalVoicePacket)?;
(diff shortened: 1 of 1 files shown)


astroftl:
Resolves #280.

astroftl:
Converting to draft while I evaluate potential impacts on RTCP packets.

astroftl:
Marked ready for review. RTCP impacts minimal. `payload_offset` points correctly to the beginning of the payload data, though this is much less useful for RTCP events. `payload_end_pad` correctly includes the nonce, however this should not be necessary since RTCP packets include a length as well as report counts which can be used to parse.

FelixMcFelix (approved):
Thanks for cleaning this up. Some of the math on those user-visible offsets had got a bit gummed up when changing over the encryption methods to the newer scheme. The start offset, at the time, was meant to be that of the extension headers. In the older encryption modes, the encryption tag was inserted right in the middle of the RTP header, before the extensions. So we *did* need to split the buffer there while we were supporting both forms of RT(C)P packet, but no longer.

Issue #245, opened 2024-08-10
`Firing Error` in a Docker container


matous-volf:
**Songbird version:** 0.4.3

**Rust version (`rustc -V`):** 1.80.1

**Serenity/Twilight version:** Serenity 0.12.2

**Output of `ffmpeg -version`, `yt-dlp --version` (if relevant):** yt-dlp 2024.08.06

**Description:**

I'm running `Call::play_only_input()` with a `YoutubeDl` source inside a Docker container. In case of interest, the code is [here](https://github.com/matous-volf/tranzistorak/blob/e4ecf73c6d9a2343210e644335184dd053daccb5/src/player/mod.rs#L132).

This is my Dockerfile (tranzistorak is the bot):

```Dockerfile
FROM rust:1.80-bullseye AS builder

RUN apt-get update
RUN apt-get install -y cmake

WORKDIR /usr/src/tranzistorak
COPY . .

RUN cargo install --path .

FROM debian:bullseye-slim AS runner

RUN apt-get update

RUN apt-get install -y python3-pip
RUN python3 -m pip install -U yt-dlp

RUN apt-get install -y libopus-dev

COPY --from=builder /usr/local/cargo/bin/tranzistorak /usr/local/bin/tranzistorak
COPY ./.env /usr/local/bin/.env

WORKDIR /usr/local/bin
CMD ["tranzistorak"]
```

When I build the image and run a container locally (Arch Linux), it works as expected with no problems:

```
INFO start:start_connection{start_shard=0 end_shard=0 total_shards=1}:initialize:boot{shard_info=[ShardId(0), ShardId(1)]}: serenity::gateway::bridge::shard_manager: Telling shard queuer to start shard 0
INFO run: serenity::gateway::bridge::shard_runner: [ShardRunner ShardInfo { id: ShardId(0), total: 1 }] Running
INFO runner{config=Config { crypto_mode: Normal, gateway_timeout: Some(10s), mix_mode: Stereo, preallocated_tracks: 1, driver_retry: Retry { strategy: Backoff(ExponentialBackoff { min: 250ms, max: 10s, jitter: 0.1 }), retry_limit: Some(5) }, use_softclip: true, driver_timeout: Some(10s), disposer: Some(DisposalThread(Sender)), scheduler: None }}: songbird::driver::connection: Connected to: frankfurt9728.discord.media:443
[...]

PR #283, opened 2025-05-22
Fix YoutubeDl by updating `yt-dlp` format syntax

mincomk:
This PR addresses [#245](https://github.com/serenity-rs/songbird/issues/245), which caused YouTube playback to fail due to recent changes in `yt-dlp` behavior.

Changed the `yt-dlp` format parameter (`-f`) to appropriate form to fix `Firing Error` occurred on `YoutubeDl` playback.

### DIFF
--- a/src/input/sources/ytdl.rs
+++ b/src/input/sources/ytdl.rs
@@ -1,11 +1,5 @@
 use crate::input::{
-    metadata::ytdl::Output,
-    AudioStream,
-    AudioStreamError,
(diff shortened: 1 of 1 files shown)


Erk-:
Is this still relevant, at least according to the man page for the latest release the old way is allowed, and is more or less equivalent to what you proposed.

https://man.archlinux.org/man/extra/yt-dlp/yt-dlp.1.en#FORMAT_SELECTION

PR #285, opened 2025-07-25
Respect the port number from VoiceServerUpdate

GnomedDev:
Discord Staff have notified bot/library developers that they will be adding voice servers soon with port numbers other than the fixed port that has been used forever, and we need to respect this.

Although the code only stripped `80` from the endpoint, let's get rid of all hard coding and just pass it in sensibly.


FelixMcFelix (approved):
Fair enough. This behaviour has been in since time immemorial at this point. The change makes sense, I think you just need to `cargo +nightly fmt`.

PR #286, opened 2025-11-23
Support Twilight 0.17.0

SuperYummy:


Erk-:
This could probably be done on current without breaking backwards compatibility.

SuperYummy:
not sure how to fix the checks. would be cool if you or someone took over

vilgotf:
The examples are still built w/ 0.16

https://github.com/serenity-rs/songbird/blob/18c282e6768c61f54c1c0b31136f8f6156aaf9b9/examples/twilight/Cargo.toml#L18-L21

Erk-:
Okay so this is pretty funky, if you have a version compiled with 0.16 in the cache it will just fail, I have tried to clear the CI caches and we can see if that resolves it.

I also made the version requirements a bit looser so 0.17.1 can get in as well.

Erk-:
The macOS test failure is unrelated (it seems to be some cmake stuff) so I will say that it is probably good now.

Erk- (approved):

Erk- (approved):

tazz4843 (approved):

PR #291, opened 2026-02-22
DAVE support

### LINKED ISSUE #293: E2EE/DAVE protocol required
**Songbird version:** 0.5

**Rust version (`rustc -V`):** rustc 1.91.1

**Serenity/Twilight version:** 0.12

**Description:**

When attempting to connect to a voice channel discord returns the error "E2EE/DAVE protocol required"

**Steps to reproduce:**
Get a songbird manager and attempt to join a voice channel using the join method

beer-psi:
Closes #293

Sending voice packets now mostly works.

Receiving is not very heavily tested, and there are a bunch of annoyances:
- ~~Passthrough (unencrypted) voice packets will still go into the decryption code path, so you will get error logs like:~~
```
ERROR davey::cryptor::frame_processors: no magic marker
```
- ~~The UDP receiving task is separate from the gateway task, so decrypting the first few packets *will* fail because the key ratchets haven't been set up yet. This should probably be fixed with a buffer?~~ Packets that are received before the key ratchets are set up are silently dropped.
- `SpeakingStateUpdate` events are apparently not fired consistently

~~The PR is still unreviewed, but~~ if your bot needs this it's been tested and working on a few bots. If you use `serenity-next`, you may want to depend on jtscuba's fork of this PR instead, which depends on `serenity-next`: https://github.com/jtscuba/songbird/tree/davey

## TODO
- [x] `serenity-voice-model` will need to be updated on crates.io to add the missing `dave_protocol_version` field in `SessionDescription`.
- [ ] Add a way to retrieve the voice privacy code and the verification codes (e.g. `Call::voice_privacy_code`?)

### DIFF
--- a/src/driver/connection/mod.rs
+++ b/src/driver/connection/mod.rs
@@ -256,7 +256,6 @@ impl Connection {
 
         spawn(ws_task::runner(interconnect.clone(), ws_state));
 
-        // TODO: Implement DAVE for receive
         #[cfg(feature = "receive")]
         spawn(udp_rx::runner(
             interconnect.clone(),
--- a/src/driver/connection/mod.rs
+++ b/src/driver/connection/mod.rs
@@ -8,15 +8,13 @@ use super::{
         message::*,
         ws::{self as ws_task, AuxNetwork},
     },
-    Config,
-    CryptoMode,
+    Config, CryptoMode,
 };
 use crate::{
     constants::*,
     model::{
         payload::{Identify, Resume, SelectProtocol},
-        Event as GatewayEvent,
-        ProtocolData,
+        Event as GatewayEvent, ProtocolData,
     },
     ws::WsStream,
     ConnectionInfo,
@@ -24,11 +22,11 @@ use crate::{
 use discortp::discord::{IpDiscoveryPacket, IpDiscoveryType, MutableIpDiscoveryPacket};
 use error::{Error, Result};
 use flume::Sender;
+use serenity_voice_model::payload::DaveMlsKeyPackage;
 use socket2::Socket;
-#[cfg(feature = "receive")]
 use std::sync::Arc;
-use std::{net::IpAddr, str::FromStr};
-use tokio::{net::UdpSocket, spawn, time::timeout};
+use std::{net::IpAddr, num::NonZeroU16, str::FromStr};
+use tokio::{net::UdpSocket, spawn, sync::RwLock, time::timeout};
 use tracing::{debug, info, instrument};
 use url::Url;
 
@@ -72,11 +70,12 @@ impl Connection {
                 session_id: info.session_id.clone(),
                 token: info.token.clone(),
                 user_id: info.user_id.into(),
+                max_dave_protocol_version: Some(davey::DAVE_PROTOCOL_VERSION),
             }))
             .await?;
 
         loop {
-            let Some(value) = client.recv_json().await? else {
+            let Some(value) = client.recv_event().await? else {
                 continue;
             };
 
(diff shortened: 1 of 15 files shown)


beer-psi:
I'm not sure how I can surface the voice privacy code up to the `Driver` interface in the existing architecture. Could I add a new `CoreEvent::VoicePrivacyCode` or something of that sort and then emit the event when a DAVE transition is executed (which is when the code is changed)?

tktcorporation:
Thanks for the great work on this! I was able to get my library working by pointing to your fork in the meantime. I may not be in a position to contribute much directly, but please let me know if there’s anything I can do to help move this forward!

tdgroot:
Works for me! 😃

LukeZurg22:
By gum, please merge this.

src/driver/tasks/ws.rs
```diff
@@ -38,6 +50,11 @@ pub(crate) struct AuxNetwork {
     attempt_idx: usize,
     info: ConnectionInfo,
 
+    dave_session: Arc<RwLock<Option<davey::DaveSession>>>,
+    dave_protocol_version: Arc<AtomicU16>,
```
Erk-, line 60: ~~This does not need to be atomic~~

Erk- (changes requested):
The changes looks good, ~~I have the comment about the unnecessary (as far as I can tell) atomic~~ (I was wrong there) and then there is some warning of unused values you get when you for example build the twilight example that needs to be cleaned up, but other than that it looks good.

Siydge:
Tested, works great for me, no issues, tested with following override for others in need

```
[patch.crates-io]
songbird = { git = "https://github.com/beerpsi-forks/songbird", branch = "davey" }
```

Erk- (approved):

Achilleees:
Thanks for this guys. Tested and works great! Looking forward to the merge.

krywicki:
Ran in to trouble with DAVE protocol on my sound bot. I've added this as a songbird patch and tested it. It  works great!  :raised_hands:

src/driver/connection/mod.rs
```diff
@@ -358,9 +375,38 @@ async fn init_cipher(
...
-                return mode
-                    .cipher_from_key(&desc.secret_key)
-                    .map_err(|_| Error::CryptoInvalidLength);
+                let dave_session =
+                    if let Some(version) = NonZeroU16::new(desc.dave_protocol_version) {
+                        let mut session = davey::DaveSession::new(
+                            version,
+                            info.user_id.0.into(),
+                            info.channel_id
+                                .expect("channel ID must be set in connection info")
```
tazz4843, line 384: If we need channel ID to establish a connection we should update this type to remove the option entirely and remove the doc string saying "channel ID is not needed" https://github.com/serenity-rs/songbird/blob/e8c134664ba4063db5a5f06d5417182d2c7514f6/src/info.rs#L106-L110

src/driver/tasks/ws.rs
```diff
@@ -38,6 +56,11 @@ pub(crate) struct AuxNetwork {
     attempt_idx: usize,
     info: ConnectionInfo,
 
+    dave_session: Arc<RwLock<Option<davey::DaveSession>>>,
```
tazz4843, line 59: Does this need to be a tokio `RwLock`? None of the methods on `DaveSession` are asynchronous, and as far as I can tell a lock is never held over an await point (if it is and I missed one, disregard this comment). For performance reasons we should prefer standard synchronous `RwLock`s if possible.

tazz4843 (changes requested):

src/driver/connection/mod.rs
beer-psi: I think there were already other `.expect`s on channel ID already so I just followed them. Not entirely sure why it wasn't just required then either.

https://github.com/serenity-rs/songbird/blob/18c282e6768c61f54c1c0b31136f8f6156aaf9b9/src/info.rs#L34-L41

src/driver/tasks/ws.rs
beer-psi: I think it was originally a Tokio RWLock because the lock was held over an await point, but a lot of things shifted since then and looks like that isn't the case anymore.

src/driver/connection/mod.rs
tazz4843: Let’s get this PR in and I can follow up with one that changes the type later.

tazz4843 (approved):

Erk- (approved):

GnomedDev (approved):

PR #294, opened 2026-03-28
Switch from audiopus to opus2

beer-psi:
Lightly tested on both send and receive. `opus2` is not compatible with `audiopus`, so this is breaking, since Songbird re-exports `audiopus`:
- `Packet` newtype is dropped and `opus2` takes a byte slice directly. An empty byte slice is used to represent packet loss. Also means that packet length checks have to be done manually.
- `Bitrate::BitsPerSecond` -> `Bitrate::Bits`
- `SampleRate` enum is dropped in favor of a simple `u32`.
- Some other module changes (e.g. `audiopus::coder::Decoder` -> `opus2::Decoder`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ version = "0.5.0"
 aead = { optional = true, version = "0.5.2" }
 aes-gcm = { optional = true, version = "0.10.3" }
 async-trait = { optional = true, version = "0.1" }
-audiopus = { optional = true, version = "0.3.0-rc.0" }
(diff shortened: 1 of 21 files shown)


Erk- (approved):
Looks good, this should also resolve build issues on some platforms right?

beer-psi:
Yeah, since it's a newer version of `libopus`. The macOS CI which failed in the DAVE PR is now good.

tazz4843 (approved):

PR #297, opened 2026-03-28
Fix low-hanging Clippy lints

tazz4843:
One Clippy lint is left over from this PR, but it's a relatively major one (too many arguments) so I left it as is.


src/driver/tasks/disposal.rs
```diff
@@ -35,6 +35,10 @@ impl DisposalThread {
 /// The mixer uses this to offload heavy and expensive drop operations
 /// to prevent deadline misses.
 #[instrument(skip(mix_rx))]
+#[allow(
```
Erk-:
```suggestion
#[expect(
```

Erk- (approved):
Just a small nit so we remember to remove it if something changes, otherwise it looks good.

PR #304, opened 2026-04-08
Allow connecting without any retries

### LINKED ISSUE #302: Not possible to attempt connection only once
**Songbird version:** (version)
0.6.0

**Rust version (`rustc -V`):** (version)
1.92.0

**Description:**
Songbird 0.6.0 switched to an `Option<NonZeroU8>` for specifying the retry limit ([docs](https://serenity-rs.github.io/songbird/current/songbird/driver/retry/struct.Retry.html#structfield.retry_limit)). As a result, it is no longer possible to specify `Some(0)` to not retry at all. Should this have been a `NonMaxU8`?

beer-psi:
Use `u8` instead of `NonZeroU8` for specifying number of retries, so `Some(0)` can be specified again to indicate that no retrying is desired.

Closes #302

### DIFF
--- a/src/driver/retry/mod.rs
+++ b/src/driver/retry/mod.rs
@@ -2,8 +2,6 @@
 
 mod strategy;
 
-use nonmax::NonMaxU8;
-
 pub use self::strategy::*;
 
 use crate::FloatDuration;
@@ -24,14 +22,14 @@ pub struct Retry {
     /// while `Some(0)` will attempt to connect *once* (no retries).
     ///
     /// *Defaults to `Some(5)`.*
-    pub retry_limit: Option<NonMaxU8>,
+    pub retry_limit: Option<u8>,
 }
 
 impl Default for Retry {
     fn default() -> Self {
         Self {
             strategy: Strategy::Backoff(ExponentialBackoff::default()),
-            retry_limit: Some(const { NonMaxU8::new(5).unwrap() }),
+            retry_limit: Some(5),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Erk-:
It is not really necessary for this to be `Non{Zero, Max}U8` since it will not change the size of Retry. So maybe it would be nice to just have it as a `u8`?

beer-psi:
Then you can't distinguish between not retrying (`Some(0)`) and infinite retries (`None`). IIRC the original type was an `Option<u8>`.

Erk-:
> Then you can't distinguish between not retrying (`Some(0)`) and infinite retries (`None`). IIRC the original type was an `Option<u8>`.

Ah yeah sorry I was unclear I meant `Option<u8>` it used to be a `Option<usize>`.

Erk- (approved):

KiloOscarSix (changes requested):
`Option<u8>` is 2 bytes long.
`Option<NonMaxU8>` would maintain the 1-byte length while allowing `Some(0)` and `None`.
`nonmax::NonMaxU8`

beer-psi:
> `Option<u8>` is 2 bytes long. `Option<NonMaxU8>` would maintain the 1-byte length while allowing `Some(0)` and `None`. `nonmax::NonMaxU8`

It doesn't matter since the size of `Retry` can't change due to alignment reasons. [Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e4fd35bd82b3ce10dd091a984153bd37)

KiloOscarSix (approved):

